PR #3, opened 2018-09-13
Pass right cq pointer to create_qp

kuenishi:
Wrong completion queue pointer passed to queue pair and somehow it works in send/recv but non-deterministic fail occurs at destroy phase, like segmentation fault, deadlock and invalid argument.


jonhoo:
Huh, that's so weird! Thanks!

PR #4, opened 2018-09-18
Add Shared Resource Queue implementation

kuenishi:
With SRQ a verbs process will be able to keep multiple QPs in non-synthetic workload. A loopback example on SRQ is added as well.

Please don't feel being rushed as I've been working on my local branch and just leaving my work as-it-is.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -455,6 +455,32 @@ impl<'a> Drop for CompletionQueue<'a> {
     }
 }
 
+/// SharedReceiveQueue, so-called srq.
+pub struct SharedReceiveQueue<'pd> {
+    _phantom: PhantomData<&'pd ()>,
+    srq: *mut ffi::ibv_srq,
+}
+
+impl<'a> Default for SharedReceiveQueue<'a> {
+    fn default() -> SharedReceiveQueue<'a> {
+        SharedReceiveQueue{
+            _phantom: PhantomData,
+            srq: ptr::null::<c_void>() as *mut _,
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -455,6 +455,32 @@ impl<'a> Drop for CompletionQueue<'a> {
...
+pub struct SharedReceiveQueue<'pd> {
+    _phantom: PhantomData<&'pd ()>,
+    srq: *mut ffi::ibv_srq,
+}
+
+impl<'a> Default for SharedReceiveQueue<'a> {
+    fn default() -> SharedReceiveQueue<'a> {
+        SharedReceiveQueue{
+            _phantom: PhantomData,
+            srq: ptr::null::<c_void>() as *mut _,
```
jonhoo: I think you can simplify this to just `srq: ptr::null(),`

src/lib.rs
```diff
@@ -518,6 +545,7 @@ impl<'res> QueuePairBuilder<'res> {
             max_send_wr,
             recv,
             max_recv_wr,
+            srq: std::option::Option::None,
```
jonhoo: This can just say `srq: None`

src/lib.rs
```diff
@@ -681,6 +709,11 @@ impl<'res> QueuePairBuilder<'res> {
         self
     }
 
+    /// Update the srq, othrewise the default NULL is used.
```
jonhoo: It'd be good to elaborate a little here on what an unset `srq` means.

kuenishi:
Thank you for the review, but it turned out we didn't need it and now I'm going to withdraw this pull request.

PR #5, opened 2018-09-18
Make QueuePairEndpoint serializable

kuenishi:
ibverbs requires another inter-process communication to exchange their endpoint for connection establishment. As QueuePairEndpoint includes private ffi pointer it is hard to serialize outside the rust-ibverbs module and it would be more natural to use standardized way of serde in Rust.


jonhoo:
Sorry that it took so long to get back to this! Why don't you just directly implement `Serialize` and `Deserialize` for `QueuePairEndpoint`?

jonhoo:
Closing in favor of #7.

PR #7, opened 2021-03-31
Implement ser/der for QueuePairEndpoint

iwahbe:
Make `QueuePairEndpoint` serializable. 

I use the same essential strategy (and some of the code) as @kuenishi. The reason we can't just slap a `#[derive(Serialize, Deserialize)]` is that `QueuePairEndpoint` contains a `C` style `union`, which `serde` does not want to mess with. I get around this by providing idiomatic conversions from/to `PortableQPEndpoint`, and telling serde to use `PortableQPEndpoint` as an intermediary for serialization and deserialization. 

For a larger struct, you could implement custom serializers and deserializers, but this copy strategy works well for small structs. 

I also implement `fmt::Debug` and `PartialEq` for `QueuePairEndpoint`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -776,11 +776,11 @@ use serde::{Deserialize, Serialize};
 /// A (serde) serializable representation of a `QueuePairEndpoint`.
 #[derive(Copy, Clone, Debug)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[doc(hidden)]
 pub struct PortableQPEndpoint {
     num: u32,
     lid: u16,
-    gid_subnet_prefix: u64,
-    gid_interface_id: u64,
+    gid_raw: [u8; 16],
 }
 
 impl From<QueuePairEndpoint> for PortableQPEndpoint {
@@ -788,9 +788,7 @@ impl From<QueuePairEndpoint> for PortableQPEndpoint {
         PortableQPEndpoint {
             num: qpe.num,
             lid: qpe.lid,
-            // TODO: they're actually be64 and fixit
-            gid_subnet_prefix: unsafe { qpe.gid.global.subnet_prefix },
-            gid_interface_id: unsafe { qpe.gid.global.interface_id },
+            gid_raw: unsafe { qpe.gid.raw },
         }
     }
 }
@@ -798,12 +796,11 @@ impl From<QueuePairEndpoint> for PortableQPEndpoint {
 impl From<PortableQPEndpoint> for QueuePairEndpoint {
     fn from(pqpe: PortableQPEndpoint) -> QueuePairEndpoint {
         let mut gid = ffi::ibv_gid::default();
-        gid.global.subnet_prefix = pqpe.gid_subnet_prefix;
-        gid.global.interface_id = pqpe.gid_interface_id;
+        gid.raw = pqpe.gid_raw;
         QueuePairEndpoint {
             num: pqpe.num,
             lid: pqpe.lid,
-            gid: gid,
+            gid,
         }
     }
 }
@@ -846,22 +843,16 @@ pub struct QueuePairEndpoint {
 
 impl PartialEq<QueuePairEndpoint> for QueuePairEndpoint {
     fn eq(&self, rhs: &QueuePairEndpoint) -> bool {
-        self.num == rhs.num
-            && self.lid == rhs.lid
-            && unsafe {
-                self.gid.global.subnet_prefix == rhs.gid.global.subnet_prefix
-                    && self.gid.global.interface_id == rhs.gid.global.interface_id
-            }
+        self.num == rhs.num && self.lid == rhs.lid && unsafe { self.gid.raw == rhs.gid.raw }
     }
 }
 
 impl std::fmt::Debug for QueuePairEndpoint {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f,"QueuePairEndpoint {{ num: {:?}, lid: {:?}, gid_subnet_prefix: {:?}, gid_interface_id: {:?} }}",
-        self.num,
-        self.lid,
-        unsafe{self.gid.global.subnet_prefix},
-        unsafe{self.gid.global.interface_id},
+        write!(
+            f,
+            "QueuePairEndpoint {{ num: {:?}, lid: {:?}, gid: union not shown }}",
+            self.num, self.lid,
         )
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -314,7 +314,7 @@ impl Context {
 
         // let mut gid = ffi::ibv_gid::default();
         let mut gid = Gid::default();
-        let ok = unsafe { ffi::ibv_query_gid(ctx, PORT_NUM, 0, gid.as_mut() as _) };
+        let ok = unsafe { ffi::ibv_query_gid(ctx, PORT_NUM, 0, gid.as_mut()) };
         if ok != 0 {
             return Err(io::Error::last_os_error());
         }
@@ -831,19 +831,13 @@ impl From<Gid> for ffi::ibv_gid {
 
 impl AsRef<ffi::ibv_gid> for Gid {
     fn as_ref(&self) -> &ffi::ibv_gid {
-        unsafe { self.raw.as_ptr().cast::<ffi::ibv_gid>().as_ref().unwrap() }
+        unsafe { &*self.raw.as_ptr().cast::<ffi::ibv_gid>() }
     }
 }
 
 impl AsMut<ffi::ibv_gid> for Gid {
     fn as_mut(&mut self) -> &mut ffi::ibv_gid {
-        unsafe {
-            self.raw
-                .as_mut_ptr()
-                .cast::<ffi::ibv_gid>()
-                .as_mut()
-                .unwrap()
-        }
+        unsafe { &mut *self.raw.as_mut_ptr().cast::<ffi::ibv_gid>() }
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -64,6 +64,7 @@
 // avoid warnings about RDMAmojo, iWARP, InfiniBand, etc. not being in backticks
 #![cfg_attr(feature = "cargo-clippy", allow(doc_markdown))]
 
+use std::convert::TryInto;
 use std::ffi::CStr;
 use std::io;
 use std::marker::PhantomData;
@@ -84,6 +85,9 @@ pub use ffi::ibv_qp_type;
 pub use ffi::ibv_wc;
 pub use ffi::ibv_wc_opcode;
 pub use ffi::ibv_wc_status;
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
 
 /// Access flags for use with `QueuePair` and `MemoryRegion`.
 pub use ffi::ibv_access_flags;
@@ -260,7 +264,7 @@ impl<'devlist> Device<'devlist> {
 pub struct Context {
     ctx: *mut ffi::ibv_context,
     port_attr: ffi::ibv_port_attr,
-    gid: ffi::ibv_gid,
+    gid: Gid,
 }
 
 unsafe impl Sync for Context {}
@@ -308,8 +312,9 @@ impl Context {
             }
         }
 
-        let mut gid = ffi::ibv_gid::default();
-        let ok = unsafe { ffi::ibv_query_gid(ctx, PORT_NUM, 0, &mut gid as *mut _) };
+        // let mut gid = ffi::ibv_gid::default();
+        let mut gid = Gid::default();
+        let ok = unsafe { ffi::ibv_query_gid(ctx, PORT_NUM, 0, gid.ffi()) };
         if ok != 0 {
             return Err(io::Error::last_os_error());
         }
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -766,16 +766,106 @@ pub struct PreparedQueuePair<'res> {
...
+// We go through this rigeramol because serde does now know how to derive unions.
+
+/// A (serde) serializable representation of a `QueuePairEndpoint`.
+#[derive(Copy, Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct PortableQPEndpoint {
+    num: u32,
+    lid: u16,
+    gid_subnet_prefix: u64,
+    gid_interface_id: u64,
```
jonhoo, line 783: This feels weird — `ibv_gid` is a union for a reason, and I'm not sure it makes sense to always interpret it as being a `global`. For example, what if the user is using the `raw` variant and encoding multiple `u16`s in it instead?

src/lib.rs
```diff
@@ -766,16 +766,106 @@ pub struct PreparedQueuePair<'res> {
...
     lid: u16,
     gid: ffi::ibv_gid,
 }
 
+impl PartialEq<QueuePairEndpoint> for QueuePairEndpoint {
+    fn eq(&self, rhs: &QueuePairEndpoint) -> bool {
+        self.num == rhs.num
+            && self.lid == rhs.lid
+            && unsafe {
+                self.gid.global.subnet_prefix == rhs.gid.global.subnet_prefix
```
jonhoo, line 852: Wouldn't it make more sense to just compare `raw` directly here?

jonhoo (commented):
Thanks! I'm a little worried about forcing the interpretation of `ibv_gid` as a `global` — `raw` is there for a reason. Is there maybe a way we can expose this in a way where the user has to assert that their `QueuePairEndpoint` is in fact using `global` before they can use it with `PortableQPEndpoint`?

src/lib.rs
```diff
@@ -766,16 +766,106 @@ pub struct PreparedQueuePair<'res> {
     rnr_retry: u8,
 }
 
+#[cfg(feature = "serde")]
+extern crate serde;
```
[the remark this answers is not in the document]
iwahbe: It isn't, but each crate needs to opt in to the 2018 edition, and I thought that was a decision you should make.

src/lib.rs
```diff
@@ -766,16 +766,106 @@ pub struct PreparedQueuePair<'res> {
...
+            }
+    }
+}
+
+impl std::fmt::Debug for QueuePairEndpoint {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f,"QueuePairEndpoint {{ num: {:?}, lid: {:?}, gid_subnet_prefix: {:?}, gid_interface_id: {:?} }}",
+        self.num,
+        self.lid,
+        unsafe{self.gid.global.subnet_prefix},
```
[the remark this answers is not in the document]
iwahbe: I agree that this looks weird and wrong, but it's what `rustfmt` produces. 

You can confirm this by running  `❯ cat lib.rs| rustfmt | rg 'unsafe\{self.gid'`. I got output:
```
        unsafe{self.gid.global.subnet_prefix},
        unsafe{self.gid.global.interface_id},
```

src/lib.rs
```diff
@@ -766,16 +766,106 @@ pub struct PreparedQueuePair<'res> {
...
+extern crate serde;
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+// We go through this rigeramol because serde does now know how to derive unions.
+
+/// A (serde) serializable representation of a `QueuePairEndpoint`.
+#[derive(Copy, Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct PortableQPEndpoint {
```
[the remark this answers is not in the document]
iwahbe: That seems right.

src/lib.rs
```diff
@@ -766,16 +766,106 @@ pub struct PreparedQueuePair<'res> {
...
+    gid_subnet_prefix: u64,
+    gid_interface_id: u64,
+}
+
+impl From<QueuePairEndpoint> for PortableQPEndpoint {
+    fn from(qpe: QueuePairEndpoint) -> PortableQPEndpoint {
+        PortableQPEndpoint {
+            num: qpe.num,
+            lid: qpe.lid,
+            // TODO: they're actually be64 and fixit
```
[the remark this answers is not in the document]
iwahbe: I said some of this code was from @kuenishi's PR, and this was. I looked for a reason for it to exist, and could not find one. I removed the comment.

src/lib.rs
iwahbe: Yes it would.

src/lib.rs
iwahbe: I think it probably makes more sense to use `ibv_gid.raw`, so I switched to that for the fixup commit. Given that it's a union, and that as far as I'm aware, any set of 128 bits is a valid `[u8; 16]`, it shouldn't matter under what type we store the union as, it just might not make sense to read. The same invariants (which value, raw or global) that needed to be upheld before the data was serialized need to be upheld when the data is deserialized.

src/lib.rs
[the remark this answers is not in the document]
jonhoo: Ah, I forgot this crate is so old it doesn't use the 2018 edition! Feel free to do that in a follow-up PR!

src/lib.rs
jonhoo: So, the issue we run into here is that if a `QueuePairEndpoint` is serialized and sent over the wire to a host with different endianness, that other host will end up using the _wrong_ numbers when accessing the fields through the subnet prefix/interface id pair. I'm not exactly sure what the right or expected behavior here is. In some sense, this question stems back to "why do you want endpoints to be serializeable?", since they sort of fundamentally aren't transferrable in a meaningful way to other hosts.

iwahbe: I'm not sure how to handle the endian problem for a union. I'm looking where to find out which discriminant is being used, but I don't see where it's encoded. 

My understanding (and I probably am wrong, if you don't understand what I'm trying to do) is that computers need to exchange endpoints over the network(non RDMA) before they can communicate over RDMA. As a `QueuePairEndpoint` is what we need to handshake on, it makes sense to just send that.

jonhoo: Ah, no, you're right, I'm just paging this all back in — it's been a while :sweat_smile: 

I'm honestly not sure what the intended function of this `union` is here. I've been reading around and information is scarce. I did find this on [this page](https://www.rdmamojo.com/2012/08/02/ibv_query_gid/):

> Every GID has 128 bits and its value is combined from two parts:
>
> - Higher 64 bits: subnet prefix. Value that identifies a set of endports which are managed by a common SM
> - Lower 64 bits: GID prefix. EUI-64 value that is configured by the SM

Which _suggests_ that the full 128 bits are actually in a well-defined format, and that we should _always_ exchange `raw` without doing any endianness business. Which then suggests that the union is only there to make it more convenient to look at these values separately. But presumably to actually read them you _may_ have to do endianness conversion (and presumably you should only ever set the entire `raw` in one go). This in turn suggests to me that we may want to implement our own `ibv_gid` that is _just_ `raw` (and not a `union`) and then (maybe) define read-only accessor methods for the subnet prefix and GID prefix that do the right endianness conversion.

In the context of this change, I think we can stick with just operating on `raw`, and the move away from `union` is just a suggestion for how we can improve the interface (by removing this ambiguity) if you have the time to give it a try :)

iwahbe: I think that https://github.com/jonhoo/rust-ibverbs/pull/7/commits/3cc8d7915c94bc229543a9752370981003a4a586 works. I replace `ibv_gid` with a rust type, and this solves our serialization problem. It makes the ffi slightly awkward, but not by much, and it clarifies how we interpret `ibv_gid`.

jonhoo: Nice, thank you! How did you figure out that big endian is the right way to read out these values? Is there a spec we can reference?

jonhoo: Nice, thank you! How did you figure out that big endian is the right way to read out these values? Is there a spec we can reference?

iwahbe: No clear spec, but I'm going off of the header files in `rdma-core`. `ibv_gid` is defined as 
```c
union ibv_gid {
  uint8_t raw[16];
  struct {
    __be64 subnet_prefix;
    __be64 interface_id;
  } global;
};
```
`__be64` is typedefed to `__bitwise __u64`. This is a [sparse](https://en.wikipedia.org/wiki/Sparse) flag. My understanding is that it informs people that it should be read as a big endian 64 bit integer, converting to a normal number with `be64_to_cpu(const __be64);`.

src/lib.rs
```diff
@@ -766,14 +771,78 @@ pub struct PreparedQueuePair<'res> {
...
+/// };
+/// ```
+///
+/// It appears that `global` exists for convenience, but can be safely ignored.
+/// For continuity, the methods `subnet_prefix` and `interface_id` are provided.
+/// These methods read the array as big endian, regardless of native cpu
+/// endianness.
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Default, Copy, Clone, Debug, Eq, PartialEq)]
+struct Gid {
```
jonhoo, line 795: This should probably be `#[repr(C)]`. Or maybe `#[repr(transparent)]`. I think either will do.

src/lib.rs
```diff
@@ -766,14 +771,78 @@ pub struct PreparedQueuePair<'res> {
...
+    raw: [u8; 16],
+}
+
+impl Gid {
+    /// Expose a pointer to the underlying `ffi::ibv_gid` object for use in ffi
+    /// calls.
+    fn ffi(&mut self) -> *mut ffi::ibv_gid {
```
jonhoo, line 802: Let's implement `AsRef<ffi::ibv_gid>` and `AsMut` this way. That _should_ mean that in general users can just pass an `&Gid` _directly_ to FFI methods, and the compiler coercion will automatically take care of it (and calling `.ffi()` _should_ be unnecessary).

src/lib.rs
jonhoo: (for my own later reference: https://github.com/linux-rdma/rdma-core/blob/6dfd71e2549c12142a1c11cfeeb1e314c2e31280/libibverbs/verbs.h#L65-L71)

Ah, that's a great catch! Yes, that looks exactly right — thanks for digging into it.

src/lib.rs
```diff
@@ -766,14 +771,78 @@ pub struct PreparedQueuePair<'res> {
...
+        self.raw.as_mut_ptr() as _
+    }
+
+    /// Expose the subnet_prefix component of the `Gid` as a u64. This is
+    /// equivalent to accessing the `global.subnet_prefix` component of the
+    /// `ffi::ibv_gid` union.
+    #[allow(dead_code)]
+    fn subnet_prefix(&self) -> u64 {
+        u64::from_be_bytes(self.raw[..8].try_into().unwrap())
```
jonhoo, line 811: It makes me really sad that we need `try_into` here. I _feel_ like it should be unnecessary, but it probably is for now.

src/lib.rs
```diff
@@ -766,14 +771,78 @@ pub struct PreparedQueuePair<'res> {
...
+///     } global;
+/// };
+/// ```
+///
+/// It appears that `global` exists for convenience, but can be safely ignored.
+/// For continuity, the methods `subnet_prefix` and `interface_id` are provided.
+/// These methods read the array as big endian, regardless of native cpu
+/// endianness.
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Default, Copy, Clone, Debug, Eq, PartialEq)]
```
jonhoo, line 794: Let's add `Hash` to this.

src/lib.rs
iwahbe: I think `#[repr(transparent)]` makes more sense, but I agree that the memory layout will be identical.

src/lib.rs
```diff
@@ -830,7 +825,25 @@ impl From<ffi::ibv_gid> for Gid {
...
 impl From<Gid> for ffi::ibv_gid {
     fn from(mut gid: Gid) -> Self {
-        unsafe { *gid.ffi() }
+        *gid.as_mut()
+    }
+}
+
+impl AsRef<ffi::ibv_gid> for Gid {
+    fn as_ref(&self) -> &ffi::ibv_gid {
+        unsafe { self.raw.as_ptr().cast::<ffi::ibv_gid>().as_ref().unwrap() }
```
jonhoo, line 834:
```suggestion
        unsafe { &*self.raw.as_ptr().cast::<ffi::ibv_gid>() }
```
I don't think we need the extra safety check that `.as_ref()` implements here since we _know_ the pointer must be non-null (as it was cast from a reference).

src/lib.rs
```diff
@@ -830,7 +825,25 @@ impl From<ffi::ibv_gid> for Gid {
...
+}
+
+impl AsMut<ffi::ibv_gid> for Gid {
+    fn as_mut(&mut self) -> &mut ffi::ibv_gid {
+        unsafe {
+            self.raw
+                .as_mut_ptr()
+                .cast::<ffi::ibv_gid>()
+                .as_mut()
+                .unwrap()
```
jonhoo, line 845:
```suggestion
            unsafe { &mut *self.raw.as_mut_ptr().cast::<ffi::ibv_gid>() }
```

src/lib.rs
```diff
@@ -830,7 +825,25 @@ impl From<ffi::ibv_gid> for Gid {
...
+        }
```
jonhoo, line 846:
```suggestion
       unsafe { &mut *self.raw.as_mut_ptr().cast::<ffi::ibv_gid>() }
```

src/lib.rs
```diff
@@ -308,8 +312,9 @@ impl Context {
             }
         }
 
-        let mut gid = ffi::ibv_gid::default();
-        let ok = unsafe { ffi::ibv_query_gid(ctx, PORT_NUM, 0, &mut gid as *mut _) };
+        // let mut gid = ffi::ibv_gid::default();
+        let mut gid = Gid::default();
+        let ok = unsafe { ffi::ibv_query_gid(ctx, PORT_NUM, 0, gid.as_mut() as _) };
```
jonhoo, line 317: I _think_ this can now just be:
```suggestion
        let ok = unsafe { ffi::ibv_query_gid(ctx, PORT_NUM, 0, &mut gid) };
```
At least that was my hope with implementing `AsRef` and `AsMut`.

iwahbe: As it turns out, you can convert `gid.as_mut() as _` to `gid.as_mut()`, but rust is not yet smart enough to figure out the types for `&mut gid`.

jonhoo (approved):
Thanks for sticking with all my requests for changes. Looks great now! :+1:

jonhoo:
Released in 0.4.5 and 0.5.0!

PR #8, opened 2021-04-09
[RFC] Separated sys crate out and applied changes from nevenoomo

rdelfin:
I've been using ibverbs for the last few months for an application I went shopping around for Rust wrappers. I found this one, but I found the lack of examples made it difficult to use and fully understand. From there I found @nevenoomo's fork together with his [netCAT](https://github.com/nevenoomo/NetCAT) application, which helped show a lot more clear examples of how to make use of this. I then forked off of that, separated out the direct bindings into a separate crate (which made it a whole lot easier to import this into a Bazel project), switched the build system to the `cmake` crate (which as an added bonus makes this work when cross-compiling) and have been working off of that branch since.

I also had to add a few minor changes to rdma-core since it has... Compilation errors if the path you build from is too nested (which seems to be an issue with the `cmake` crate).

This PR is by no means ready to merge, but I wanted to get some feedback on whether this change would even be palatable. If so, I can work on cleaning up further. Let me know what you think!


jonhoo:
Hi there!

This is very exciting stuff, thanks for pushing on it! I'll take a read through and leave some more specific comments. It may be worthwhile splitting this into multiple smaller PRs just to make it easier to review, but overall I like the directions!

.gitignore
```diff
@@ -1,3 +1,5 @@
 target/
 **/*.rs.bk
 Cargo.lock
+vendor/bindgen
```
jonhoo, line 4: You can just remove this one entirely now (see latest `master`).

.gitmodules
```diff
@@ -1,6 +1,3 @@
 [submodule "vendor/rdma-core"]
-	path = vendor/rdma-core
-	url = https://github.com/linux-rdma/rdma-core.git
-[submodule "vendor/bindgen"]
-	path = vendor/bindgen
-	url = https://github.com/servo/rust-bindgen.git
+	path = ibverbs-sys/vendor/rdma-core
+	url = https://github.com/rdelfin/rdma-core.git
```
jonhoo, line 3: Hmm, why did you switch which repository this is using? I'd prefer to continue using the official one.

.travis.yml
```diff
@@ -18,7 +18,6 @@ addons:
     - george-edison55-precise-backports # cmake 3.2.3
 env:
   - LLVM_VERSION=3.9.0
-before_install: . vendor/bindgen/ci/before_install.sh
```
jonhoo, line 21: I've since removed this from `master` :)

ibverbs-sys/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+build = "build.rs"
+links = "ibverbs"
+
+keywords = ["rdma", "ibverbs", "networking", "ffi"]
+categories = ["network-programming", "api-bindings"]
+
+license = "MIT/Apache-2.0"
+
+[build-dependencies]
+bindgen = "0.57"
```
jonhoo, line 19: Let's do `0.58` while we're at it ;)

ibverbs-sys/build.rs
```diff
@@ -10,33 +11,47 @@ fn main() {
...
 
     // build vendor/rdma-core
-    Command::new("bash")
-        .current_dir("vendor/rdma-core/")
-        .args(&["build.sh"])
-        .status()
-        .expect("Failed to build vendor/rdma-core using build.sh");
+    let dst = cmake::Config::new("vendor/rdma-core")
+        .define("NO_MAN_PAGES", "1")
+        .build();
```
jonhoo, line 29: I keep wondering whether we should even build `rdma-core` as part of building this crate. Ideally we'd _just_ generate the bindings, and just require that the user already have `libibverbs` available from _somewhere_. What do you think?

ibverbs-sys/src/lib.rs
```diff
@@ -48,7 +50,6 @@ pub struct ibv_wc {
     pub dlid_path_bits: u8,
 }
 
-#[cfg_attr(feature = "cargo-clippy", allow(len_without_is_empty))]
```
jonhoo, line 51: Does clippy not warn on that any more? I think this should probably be
```rust
#[allow(clippy::len_without_is_empty)]
```
these days.

ibverbs/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+documentation = "https://docs.rs/ibverbs/"
+
+keywords = ["rdma", "ibverbs", "networking", "ffi"]
+categories = ["network-programming", "api-bindings"]
+
+license = "MIT/Apache-2.0"
+
+[badges]
+travis-ci = { repository = "jonhoo/rust-ibverbs" }
+maintenance = { status = "looking-for-maintainer" }
```
jonhoo, line 24: _nudge nudge_

ibverbs-sys/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+cmake = "0.1"
+serde = {version = "1.0.104", features = ["derive"] }
```
jonhoo, line 21: This should _probably_ just be `= "1"`.

ibverbs/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+
+[dependencies]
+serde = {version = "1.0.104", features = ["derive"] }
```
jonhoo, line 27: This should _probably_ just be `= "1"`.

ibverbs/src/lib.rs
```diff
@@ -60,26 +60,19 @@
 //! [RDMAmojo]: http://www.rdmamojo.com/
 //! [1]: http://www.rdmamojo.com/2012/05/18/libibverbs/
 
-#![deny(missing_docs)]
```
jonhoo, line 63: I'd like to keep this.

ibverbs/src/lib.rs
```diff
@@ -60,26 +60,19 @@
...
 // avoid warnings about RDMAmojo, iWARP, InfiniBand, etc. not being in backticks
-#![cfg_attr(feature = "cargo-clippy", allow(doc_markdown))]
+extern crate serde;
```
jonhoo, line 64: This should no longer be needed now that the crate is on the 2018 edition.

ibverbs/src/lib.rs
```diff
@@ -199,7 +192,7 @@ impl<'devlist> Device<'devlist> {
     ///  - `ENOMEM`: Out of memory (from `ibv_query_port_attr`).
     ///  - `EMFILE`: Too many files are opened by this process (from `ibv_query_gid`).
     ///  - Other: the device is not in `ACTIVE` or `ARMED` state.
-    pub fn open(&self) -> io::Result<Context> {
+    pub fn open(&self) -> io::Result<Arc<Context>> {
```
jonhoo, line 195: This change is suspicious to me. I don't think we should mandate that callers use an `Arc`. It's true that it makes the lifetimes go away since we can ensure that the `Context` lives for as long as we want, but does having the lifetime constraints actually post a problem for writing real code? Just having the `Context` live for "however long it needs to" _feels_ wrong somehow.

ibverbs/src/lib.rs
```diff
@@ -368,33 +372,53 @@ impl Context {
...
         } else {
-            Ok(ProtectionDomain { ctx: self, pd })
+            Ok(ProtectionDomain {
+                ctx: self.clone(),
+                pd,
+            })
+        }
+    }
+
+    /// Query corresponding device
```
jonhoo, line 387: This could definitely use some more docs :)

ibverbs/src/lib.rs
```diff
@@ -256,6 +249,9 @@ impl<'devlist> Device<'devlist> {
     }
 }
 
+/// RDMA device attributes
+pub type DeviceAttr = ffi::ibv_device_attr;
```
jonhoo, line 253: Should we wrap this type to provide nice accessors instead of just dumping out the `ffi` type directly?

ibverbs/src/lib.rs
```diff
@@ -368,33 +372,53 @@ impl Context {
...
 
 impl Drop for Context {
     fn drop(&mut self) {
-        let ok = unsafe { ffi::ibv_close_device(self.ctx) };
+        let ok = if !self.ctx.is_null() {
+            unsafe { ffi::ibv_close_device(self.ctx) }
+        } else {
+            0
+        };
         assert_eq!(ok, 0);
```
jonhoo, line 408:
```suggestion
        if !self.ctx.is_null() {
            let ok = unsafe { ffi::ibv_close_device(self.ctx) };
            assert_eq!(ok, 0);
        }
```

ibverbs/src/lib.rs
```diff
@@ -368,33 +372,53 @@ impl Context {
...
         }
+
+        Ok(dev_attrs)
     }
 }
 
 impl Drop for Context {
     fn drop(&mut self) {
-        let ok = unsafe { ffi::ibv_close_device(self.ctx) };
+        let ok = if !self.ctx.is_null() {
```
jonhoo, line 403: How can `ctx` even be NULL here? Wouldn't that mean the context failed to be constructed in the first place?

ibverbs/src/lib.rs
```diff
@@ -368,33 +372,53 @@ impl Context {
...
-    _phantom: PhantomData<&'ctx ()>,
+pub struct CompletionQueue {
+    // _phantom: PhantomData<&'ctx ()>,
     cq: *mut ffi::ibv_cq,
 }
 
-unsafe impl<'a> Send for CompletionQueue<'a> {}
-unsafe impl<'a> Sync for CompletionQueue<'a> {}
+unsafe impl Send for CompletionQueue {}
+unsafe impl Sync for CompletionQueue {}
```
jonhoo, line 419:
```suggestion
unsafe impl Send for CompletionQueue<'_> {}
unsafe impl Sync for CompletionQueue<'_> {}
```

ibverbs/src/lib.rs
```diff
@@ -368,33 +372,53 @@ impl Context {
...
+        };
         assert_eq!(ok, 0);
     }
 }
 
 /// A completion queue that allows subscribing to the completion of queued sends and receives.
-pub struct CompletionQueue<'ctx> {
-    _phantom: PhantomData<&'ctx ()>,
+pub struct CompletionQueue {
+    // _phantom: PhantomData<&'ctx ()>,
```
jonhoo, line 414: I was under the impression that a CQ _must_ outlive the context — is that not right?

ibverbs/src/lib.rs
```diff
@@ -444,12 +468,20 @@ impl<'ctx> CompletionQueue<'ctx> {
     }
 }
 
-impl<'a> Drop for CompletionQueue<'a> {
+impl Drop for CompletionQueue {
```
jonhoo, line 471:
```suggestion
impl Drop for CompletionQueue<'_> {
```

ibverbs/src/lib.rs
```diff
@@ -444,12 +468,20 @@ impl<'ctx> CompletionQueue<'ctx> {
...
     fn drop(&mut self) {
-        let errno = unsafe { ffi::ibv_destroy_cq(self.cq) };
+        let errno = if !self.cq.is_null() {
+            unsafe { ffi::ibv_destroy_cq(self.cq) }
+        } else {
+            0
+        };
```
jonhoo, line 477:
```suggestion
        if self.cq.is_null() {
            return;
        }
        let errno = unsafe { ffi::ibv_destroy_cq(self.cq) };
```

ibverbs/src/lib.rs
```diff
@@ -444,12 +468,20 @@ impl<'ctx> CompletionQueue<'ctx> {
     }
 }
 
-impl<'a> Drop for CompletionQueue<'a> {
+impl Drop for CompletionQueue {
     fn drop(&mut self) {
-        let errno = unsafe { ffi::ibv_destroy_cq(self.cq) };
+        let errno = if !self.cq.is_null() {
```
jonhoo, line 473: Here, too, how can `self.cq` be NULL — that should make the constructor fail.

ibverbs/src/lib.rs
```diff
@@ -769,14 +804,57 @@ pub struct PreparedQueuePair<'res> {
 /// An identifier for the network endpoint of a `QueuePair`.
 ///
 /// Internally, this contains the `QueuePair`'s `qp_num`, as well as the context's `lid` and `gid`.
-#[derive(Copy, Clone)]
+#[derive(Clone)]
```
jonhoo, line 807: Why remove `Copy` from this type?

ibverbs/src/lib.rs
```diff
@@ -769,14 +804,57 @@ pub struct PreparedQueuePair<'res> {
...
     gid: ffi::ibv_gid,
 }
 
-impl<'res> PreparedQueuePair<'res> {
+use serde::{Deserialize, Serialize};
+
+/// A wrapper around `QueuePairEndpoint` with addintional `rkey` and `raddr` field. Intended to be sent to the peer.
+// TODO: write more thoughout documentation for fields.
+#[derive(Serialize, Deserialize, Clone)]
+pub struct EndpointMsg {
```
jonhoo, line 819: We can probably make this nicer with the changes that landed in #7.

ibverbs/src/lib.rs
```diff
@@ -918,31 +1042,48 @@ impl<T> MemoryRegion<T> {
...
+pub struct RemoteAddr(pub u64);
 
 impl<T> Drop for MemoryRegion<T> {
     fn drop(&mut self) {
-        let errno = unsafe { ffi::ibv_dereg_mr(self.mr) };
+        let errno = if !self.mr.is_null() {
+            unsafe { ffi::ibv_dereg_mr(self.mr) }
+        } else {
+            0
+        };
```
jonhoo, line 1065:
```suggestion
        if self.mr.is_null() {
            return;
        }
        let errno = unsafe { ffi::ibv_dereg_mr(self.mr) };
```

ibverbs/src/lib.rs
```diff
@@ -918,31 +1042,48 @@ impl<T> MemoryRegion<T> {
...
+pub struct RemoteKey(pub u32);
+
+/// A remote address to perform RDMA operations on.
+#[derive(Copy, Clone, Deserialize, Serialize)]
+pub struct RemoteAddr(pub u64);
 
 impl<T> Drop for MemoryRegion<T> {
     fn drop(&mut self) {
-        let errno = unsafe { ffi::ibv_dereg_mr(self.mr) };
+        let errno = if !self.mr.is_null() {
```
jonhoo, line 1061: And here as well, it feels like NULL should be handled by the constructor.

ibverbs/src/lib.rs
```diff
@@ -1019,7 +1160,7 @@ impl<'ctx> ProtectionDomain<'ctx> {
             ffi::ibv_reg_mr(
                 self.pd,
                 data.as_mut_ptr() as *mut _,
-                n * mem::size_of::<T>(),
+                (n * mem::size_of::<T>()) as u64,
```
jonhoo, line 1163: This shouldn't be necessary on `master`.

ibverbs/src/lib.rs
```diff
@@ -1035,17 +1176,25 @@ impl<'ctx> ProtectionDomain<'ctx> {
...
 
-impl<'a> Drop for ProtectionDomain<'a> {
+impl Drop for ProtectionDomain {
     fn drop(&mut self) {
-        let errno = unsafe { ffi::ibv_dealloc_pd(self.pd) };
+        let errno = if !self.pd.is_null() {
+            unsafe { ffi::ibv_dealloc_pd(self.pd) }
+        } else {
+            0
+        };
```
jonhoo, line 1190:
```suggestion
        if self.pd.is_null() {
            return;
        }
        let errno = unsafe { ffi::ibv_dealloc_pd(self.pd) };
```

ibverbs/src/lib.rs
```diff
@@ -1035,17 +1176,25 @@ impl<'ctx> ProtectionDomain<'ctx> {
...
+            Ok(MemoryRegion { mr, size: n, data })
         }
     }
 }
 
-impl<'a> Drop for ProtectionDomain<'a> {
+impl Drop for ProtectionDomain {
     fn drop(&mut self) {
-        let errno = unsafe { ffi::ibv_dealloc_pd(self.pd) };
+        let errno = if !self.pd.is_null() {
```
jonhoo, line 1186: And same re NULL

ibverbs/src/lib.rs
```diff
@@ -1057,15 +1206,24 @@ impl<'a> Drop for ProtectionDomain<'a> {
...
-pub struct QueuePair<'res> {
-    _phantom: PhantomData<&'res ()>,
+pub struct QueuePair {
+    // _phantom: PhantomData<&'res ()>,
     qp: *mut ffi::ibv_qp,
 }
 
-unsafe impl<'a> Send for QueuePair<'a> {}
-unsafe impl<'a> Sync for QueuePair<'a> {}
+impl Default for QueuePair {
```
jonhoo, line 1214: What's the motivation for implementing `Default` for `QueuePair`?

ibverbs/src/lib.rs
```diff
@@ -1226,15 +1384,173 @@ impl<'res> QueuePair<'res> {
             Ok(())
         }
     }
+
+    /// Write N values from the specified MR to the remote memory.
```
jonhoo, line 1388: This could use some more documentation.

ibverbs/src/lib.rs
```diff
@@ -1226,15 +1384,173 @@ impl<'res> QueuePair<'res> {
...
+            Err(io::Error::new(
+                io::ErrorKind::Other,
+                format!("ERROR: Faile to write. errno={}", errno),
+            ))
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Write a single value from the top of the memory region
```
jonhoo, line 1450: This could use some more documentation. Especially with regards to why it is `unsafe`.

ibverbs/src/lib.rs
```diff
@@ -1226,15 +1384,173 @@ impl<'res> QueuePair<'res> {
...
+            Err(io::Error::new(
+                io::ErrorKind::Other,
+                format!("ERROR: Faile to read. errno={}", errno),
+            ))
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Write a single value from the top of the memory region
```
jonhoo, line 1512: This could use some more documentation. Especially with regards to why it is `unsafe`.

ibverbs/src/lib.rs
```diff
@@ -1226,15 +1384,173 @@ impl<'res> QueuePair<'res> {
...
+        mr: &MemoryRegion<T>,
+        addr: u64,
+        key: u32,
+        wr_id: u64,
+        sig: bool,
+    ) -> io::Result<()> {
+        self.post_write_buf(mr, 1, addr, key, wr_id, sig)
+    }
+
+    /// Write a single value from the top of the memory region
```
jonhoo, line 1525: This could use some more documentation. Especially with regards to why it is `unsafe`.

ibverbs/src/lib.rs
```diff
@@ -1226,15 +1384,173 @@ impl<'res> QueuePair<'res> {
...
+                format!("ERROR: Faile to read. errno={}", errno),
+            ))
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Write a single value from the top of the memory region
+    #[inline]
+    pub unsafe fn post_write_single<T>(
```
jonhoo, line 1514: What's the reason for having this helper method, given that it's longer to type out than just calling the non-single variant with an extra argument of 1?

ibverbs/src/lib.rs
```diff
@@ -1226,15 +1384,173 @@ impl<'res> QueuePair<'res> {
...
+    #[inline]
+    pub unsafe fn post_read_single<T>(
+        &self,
+        mr: &MemoryRegion<T>,
+        addr: u64,
+        key: u32,
+        wr_id: u64,
+        sig: bool,
+    ) -> io::Result<()> {
+        self.post_read_buf(mr, 1, addr, key, wr_id, sig)
```
jonhoo, line 1535: What's the reason for having this helper method, given that it's longer to type out than just calling the non-single variant with an extra argument of 1?

ibverbs/src/lib.rs
```diff
@@ -1226,15 +1384,173 @@ impl<'res> QueuePair<'res> {
...
-impl<'a> Drop for QueuePair<'a> {
+impl Drop for QueuePair {
     fn drop(&mut self) {
         // TODO: ibv_destroy_qp() fails if the QP is attached to a multicast group.
-        let errno = unsafe { ffi::ibv_destroy_qp(self.qp) };
+        let errno = if !self.qp.is_null() {
+            unsafe { ffi::ibv_destroy_qp(self.qp) }
+        } else {
+            0
+        };
```
jonhoo, line 1546:
```suggestion
        if self.qp.is_null() {
            return;
        }
        let errno = unsafe { ffi::ibv_destroy_qp(self.qp) };
```

ibverbs/src/lib.rs
```diff
@@ -1226,15 +1384,173 @@ impl<'res> QueuePair<'res> {
...
+        self.post_read_buf(mr, 1, addr, key, wr_id, sig)
+    }
 }
 
-impl<'a> Drop for QueuePair<'a> {
+impl Drop for QueuePair {
     fn drop(&mut self) {
         // TODO: ibv_destroy_qp() fails if the QP is attached to a multicast group.
-        let errno = unsafe { ffi::ibv_destroy_qp(self.qp) };
+        let errno = if !self.qp.is_null() {
```
jonhoo, line 1542: Here I think the value can be NULL if we want `QueuePair` to implement `Default`, but I'd still like to know what the use-case for that is. If there isn't one, we should instead enforce in the constructor that QP is never NULL.

ibverbs/README.md
```diff
@@ -0,0 +1,67 @@
+# ibverbs
```
jonhoo, line 1: We probably still want a top-level `README` as well. Not sure how other projects go about this? Maybe symlink one to the other?

ibverbs-sys/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
```
jonhoo, line 1: `ibverbs-sys` should also get a short `README` that points to `ibverbs`.

.travis.yml
rdelfin: Nice! Rebase it is

.gitmodules
rdelfin: This was mostly since it seems like they have compiler errors in clang (or at least, in directories that are nested enough) but I'll see if there's a better solution

ibverbs-sys/Cargo.toml
rdelfin: Good lord has it been that long since I started it?

ibverbs-sys/build.rs
rdelfin: That's a good question. Let me look around and see what other sys crates do. Vendoring is definitely nice since it means you don't need to pull in external sources. That said, making it mandatory is not great. What I can do is make it a feature flag and only used a vendored version (from a tarball or something) if the flag is enabled.

ibverbs-sys/src/lib.rs
rdelfin: Oh interesting... This might be part of a bad rebase

ibverbs/Cargo.toml
rdelfin: Oh right! I was gonna ask about that. I'd honestly be glad to help maintain!

ibverbs/src/lib.rs
rdelfin: Makes sense. Let me see how to clean up

ibverbs/src/lib.rs
rdelfin: That's a fair point. I'll look into making the lifetimes more explicit instead

ibverbs/src/lib.rs
rdelfin: Sounds like a good idea, given that's what the other structs do :D

ibverbs/src/lib.rs
rdelfin: This one I'm not 100% sure on. Let me see if I can find the original commit and whether that explains it

ibverbs/src/lib.rs
rdelfin: Let me double check this one too. The changes on this file are mostly from nevenmoo, so this might just have been them getting around lifetime requirements for ease of use.

ibverbs/src/lib.rs
rdelfin: Oh! That's some good timing

ibverbs/README.md
rdelfin: Let me look around how projects with workspaces handle this. IIRC they usually just have identical readme's on the root and main crate, but a symlink might be nicer.

rdelfin:
By the looks of it, the biggest single thing to send out in a separate PR is the `sys` vs regular crate change. I'll send out a separate PR with just that first  and leave this one open as a reference mostly

.gitmodules
jonhoo: You could try sending the patches upstream? Always good if we can help the greater ecosystem!

ibverbs/Cargo.toml
jonhoo: That's awesome! Let's work our way through this PR, and then chat about it afterwards? This crate is fairly low maintenance in the first place, but it could certainly use some love (like this PR) :)

ibverbs/src/lib.rs
jonhoo: I think they were all correct the way they were, so I'm mostly curious why the decision was made to remove them. Was there a particular use that became too cumbersome?

ibverbs/src/lib.rs
jonhoo: Should we tag them in to this PR? They may be able to provide some additional input.

jonhoo:
That sounds like a great plan — I think that is indeed a very neatly separable change.

ibverbs/src/lib.rs
rdelfin: My assumption is that the original author removed this out of ease of use. Looking through it I should have reviewed this whole file more carefully, since the one big important thing added were the device attributes

ibverbs/src/lib.rs
rdelfin: Oh! I did, it's @nevenoomo. With another tag they might see it :P

rdelfin:
I'm really sorry for taking so long on this. I've been a bit swamped with work (ironically using this library!) and forgot to get back to this.
@jonhoo am I reading correctly that a lot of the changes I put in here have already made it to master, but with correct lifetimes instead of arcs everywhere?

jonhoo:
No worries at all — open-source is basically the act of trying to organize lots of people's volunteer time :sweat_smile: 

> Am I reading correctly that a lot of the changes I put in here have already made it to master, but with correct lifetimes instead of arcs everywhere?

Hard for me to say. Some definitely have, but I don't know if _all_ of them have. You'd have to walk through the diff and see unfortunately.

rdelfin:
No worries, let me go through it and see what is missing. I did separate out the split creates too!

PR #10, opened 2021-04-17
bump to rdma-core v34.0

daniel-noland:
Update to rdma-core version 34.0 (current release).

All that was required was to correct a single typing issue to compile on
rdma-core v34.0


jonhoo (approved):
Excellent, thank you!

PR #11, opened 2021-05-31
Fix version mistake in Cargo.toml

daniel-noland:
The version number was bumped to 0.6 in version control with the last commit.  It seems that we did not reflect this change in the Cargo.toml file as well.


jonhoo:
Ah, I just forgot to push, my bad :sweat_smile:

daniel-noland:
> Ah, I just forgot to push, my bad

Cool.  Thanks for getting it sorted.

PR #16, opened 2021-08-22
[WIP] Add Guid type hiding raw ibv GUID

planetA:
Create new Guid type that takes care of endianness conversion.

### DIFF
--- a/ibverbs/src/lib.rs
+++ b/ibverbs/src/lib.rs
@@ -198,7 +198,7 @@ pub struct Guid {
 impl Guid {
     /// Upper 24 bits of the GUID are OUI (Organizationally Unique Identifier,
     /// http://standards-oui.ieee.org/oui/oui.txt). The function returns OUI as
-    /// a 24-bit number inside a u64 number.
+    /// a 24-bit number inside a u32.
     pub fn oui(&self) -> u32 {
         let padded = [0, self.raw[0], self.raw[1], self.raw[2]];
         u32::from_be_bytes(padded)
--- a/ibverbs/src/lib.rs
+++ b/ibverbs/src/lib.rs
@@ -204,8 +204,7 @@ impl Guid {
         u32::from_be_bytes(padded)
     }
 
-    /// A GUID equal to zero is considered to be reserved. Return if the GUID is
-    /// reserved.
+    /// Returns `true` if this GUID is all zeroes, which is considered reserved.
     pub fn is_reserved(&self) -> bool {
         self.raw == [0; 8]
     }
--- a/ibverbs/src/lib.rs
+++ b/ibverbs/src/lib.rs
@@ -183,6 +183,54 @@ impl<'d> From<&'d *mut ffi::ibv_device> for Device<'d> {
     }
 }
 
+/// A Global unique identifier for ibv.
+///
+/// This struct acts as a rust wrapper for GUID value represented as `__be64` in
+/// libibverbs. We introduce this struct, because u64 is stored in host
+/// endianness, whereas ibverbs stores GUID in network order (big endian).
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Default, Copy, Clone, Debug, Eq, PartialEq, Hash)]
+#[repr(transparent)]
+pub struct Guid {
+    raw: [u8; 8],
+}
+
+impl Guid {
+    /// Upper 24 bits of the GUID are OUI (Organizationally Unique Identifier,
+    /// http://standards-oui.ieee.org/oui/oui.txt). The function returns OUI as
+    /// a 24-bit number inside a u64 number.
+    pub fn oui(&self) -> u32 {
+        let padded = [0, self.raw[0], self.raw[1], self.raw[2]];
+        u32::from_be_bytes(padded)
+    }
+
+    /// A GUID equal to zero is considered to be reserved. Return if the GUID is
+    /// reserved.
+    pub fn is_reserved(&self) -> bool {
+        self.raw == [0; 8]
+    }
+}
+
+impl From<u64> for Guid {
+    fn from(guid: u64) -> Self {
+        Self {
+            raw: guid.to_be_bytes(),
+        }
+    }
+}
+
+impl From<Guid> for u64 {
+    fn from(guid: Guid) -> Self {
+        u64::from_be_bytes(guid.raw)
+    }
+}
+
+impl AsRef<ffi::__be64> for Guid {
+    fn as_ref(&self) -> &ffi::__be64 {
+        unsafe { &*self.raw.as_ptr().cast::<ffi::__be64>() }
+    }
+}
+
 impl<'devlist> Device<'devlist> {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -189,6 +189,76 @@ impl<'d> From<&'d *mut ffi::ibv_device> for Device<'d> {
...
+            0,
+            0,
+            self.0[0],
+            self.0[1],
+            self.0[2],
+        ];
+        u64::from_be_bytes(padded.try_into().unwrap())
+    }
+
+    fn is_nil(&self) -> bool {
```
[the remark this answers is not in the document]
planetA: IBTA 4.1.1 6) says:

> The unicast GID address 0:0:0:0:0:0:0:0 is reserved - referred to as
> the Reserved GID. It shall never be assigned to any endport.

src/lib.rs
```diff
@@ -189,6 +189,76 @@ impl<'d> From<&'d *mut ffi::ibv_device> for Device<'d> {
...
+#[repr(transparent)]
+pub struct Guid ([u8; 8]);
+
+impl Guid {
+    /// Upper 24 bits of the GUID are OUI (Organizationally Unique Identifier,
+    /// http://standards-oui.ieee.org/oui/oui.txt). The function returns OUI as
+    /// a 24-bit number inside a u64 number.
+    #[allow(dead_code)]
+    fn oui(&self) -> u64 {
+        let padded = vec![
```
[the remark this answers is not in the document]
jonhoo: `u32` seems like the right choice, yeah!

src/lib.rs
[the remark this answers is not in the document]
jonhoo: In that case, how about calling this `is_reserved`, with documentation that includes that quote from the spec?

src/lib.rs
```diff
@@ -189,6 +189,64 @@ impl<'d> From<&'d *mut ffi::ibv_device> for Device<'d> {
...
+    /// a 24-bit number inside a u64 number.
+    #[allow(dead_code)]
+    fn oui(&self) -> u32 {
+        let padded = vec![
+            0,
+            self.raw[0],
+            self.raw[1],
+            self.raw[2],
+        ];
+        u32::from_be_bytes(padded.try_into().unwrap())
```
jonhoo, line 216: This should not be a `vec!`, but instead just an array directly:
```suggestion
        let padded = [
            0,
            self.raw[0],
            self.raw[1],
            self.raw[2],
        ];
        u32::from_be_bytes(padded)
```

src/lib.rs
```diff
@@ -189,6 +189,64 @@ impl<'d> From<&'d *mut ffi::ibv_device> for Device<'d> {
...
+pub struct Guid {
+    raw: [u8; 8],
+}
+
+impl Guid {
+    /// Upper 24 bits of the GUID are OUI (Organizationally Unique Identifier,
+    /// http://standards-oui.ieee.org/oui/oui.txt). The function returns OUI as
+    /// a 24-bit number inside a u64 number.
+    #[allow(dead_code)]
+    fn oui(&self) -> u32 {
```
jonhoo, line 209: This should be `pub`, otherwise callers can't access it. That will also let you remove the `#[allow(dead_code)]`.

src/lib.rs
```diff
@@ -189,6 +189,64 @@ impl<'d> From<&'d *mut ffi::ibv_device> for Device<'d> {
...
+        let padded = vec![
+            0,
+            self.raw[0],
+            self.raw[1],
+            self.raw[2],
+        ];
+        u32::from_be_bytes(padded.try_into().unwrap())
+    }
+
+    fn is_reserved(&self) -> bool {
```
jonhoo, line 219:
```suggestion
    pub fn is_reserved(&self) -> bool {
```

jonhoo: It'll also need some documentation

src/lib.rs
```diff
@@ -189,6 +189,64 @@ impl<'d> From<&'d *mut ffi::ibv_device> for Device<'d> {
...
+        self.raw.iter().all(|i| *i == 0_u8)
```
jonhoo, line 220: Does
```suggestion
        self.raw == [0; 8]
```
work?

src/lib.rs
```diff
@@ -189,6 +189,64 @@ impl<'d> From<&'d *mut ffi::ibv_device> for Device<'d> {
...
+    fn from(guid: u64) -> Self {
+        Self {
+            raw: guid.to_be_bytes(),
+        }
+    }
+}
+
+impl From<Guid> for u64 {
+    fn from(guid: Guid) -> Self {
+        u64::from_be_bytes(guid.raw[..].try_into().unwrap())
```
jonhoo, line 234: Does this not work?
```suggestion
        u64::from_be_bytes(guid.raw)
```

src/lib.rs
```diff
@@ -189,6 +189,64 @@ impl<'d> From<&'d *mut ffi::ibv_device> for Device<'d> {
...
+        unsafe { &*self.raw.as_ptr().cast::<ffi::__be64>() }
+    }
+}
+
+impl AsMut<ffi::__be64> for Guid {
+    fn as_mut(&mut self) -> &mut ffi::__be64 {
+        unsafe { &mut *self.raw.as_mut_ptr().cast::<ffi::__be64>() }
+    }
+}
+
```
jonhoo, line 249: Why is this impl needed?

jonhoo (changes requested):
Looking better! I left a few more inline comments.

ibverbs/src/lib.rs
```diff
@@ -183,6 +183,54 @@ impl<'d> From<&'d *mut ffi::ibv_device> for Device<'d> {
...
+#[derive(Default, Copy, Clone, Debug, Eq, PartialEq, Hash)]
+#[repr(transparent)]
+pub struct Guid {
+    raw: [u8; 8],
+}
+
+impl Guid {
+    /// Upper 24 bits of the GUID are OUI (Organizationally Unique Identifier,
+    /// http://standards-oui.ieee.org/oui/oui.txt). The function returns OUI as
+    /// a 24-bit number inside a u64 number.
```
jonhoo, line 201:
```suggestion
    /// a 24-bit number inside a u32.
```
I think this is what you mean?

ibverbs/src/lib.rs
```diff
@@ -183,6 +183,54 @@ impl<'d> From<&'d *mut ffi::ibv_device> for Device<'d> {
...
+    pub fn oui(&self) -> u32 {
+        let padded = [0, self.raw[0], self.raw[1], self.raw[2]];
+        u32::from_be_bytes(padded)
+    }
+
+    /// A GUID equal to zero is considered to be reserved. Return if the GUID is
+    /// reserved.
```
jonhoo, line 208:
```suggestion
    /// Returns `true` if this GUID is all zeroes, which is considered reserved.
```

jonhoo (approved):

jonhoo:
Thanks!

PR #20, opened 2021-11-04
Split out an ibverbs-sys crate

rdelfin:
This change does some cleanup and splits out the ffi bindings into their own ibverbs-sys crates.

This makes it easier for someone to use the ffi bindings by themselves, or to potentially override the bindings in a build system like, say, Bazel, where build.rs can't be easily executed and used.


ibverbs-sys/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "ibverbs-sys"
+version = "0.1.0"
```
rdelfin, line 3: Wasn't sure whether to use the same version number as `ibverbs` or to start counting from scratch so they're "separate". Up to you!

ibverbs/src/lib.rs
```diff
@@ -267,7 +261,10 @@ impl<'devlist> Device<'devlist> {
     pub fn index(&self) -> io::Result<i32> {
         let idx = unsafe { ffi::ibv_get_device_index(*self.0) };
         if idx == -1 {
-            Err(io::Error::new(io::ErrorKind::Unsupported, "device index not known"))
+            Err(io::Error::new(
+                io::ErrorKind::Unsupported,
+                "device index not known",
+            ))
```
rdelfin, line 267: Is it possible I have different rustfmt settings? And should these be specified in a config at the top of the repo?

ibverbs-sys/Cargo.toml
jonhoo: I think keeping them separate makes a lot of sense — that's almost certainly what we want anyway, so that we can cut releases independently for the two.

ibverbs/src/lib.rs
jonhoo: You probably have a newer version of rustfmt than I had when I wrote that. Totally fine :)

jonhoo (approved):
It's a good refactor. Will publish this today!

PR #26, opened 2024-01-18
Make the Endpoint fields accessible and the Gid optional

YtvwlD:
I do have an existing C application I want to communicate with and I think it's not using global routing, so I have to get the (lid, qpn) information out of this library and into the other application (and back).

Because `QueuePairEndpoint` is already `Serialize` and `Deserialize`, I don't think that making the fields public has any serious implications (though adding getter functions and a `new` function might be nicer, idk).

### DIFF
--- a/ibverbs/src/lib.rs
+++ b/ibverbs/src/lib.rs
@@ -919,7 +919,7 @@ pub struct QueuePairEndpoint {
     /// the context's `lid`
     pub lid: u16,
     /// the context's `gid`, used for global routing
-    pub gid: Gid,
+    pub gid: Option<Gid>,
 }
 
 impl<'res> PreparedQueuePair<'res> {
@@ -932,7 +932,7 @@ impl<'res> PreparedQueuePair<'res> {
         QueuePairEndpoint {
             num,
             lid: self.ctx.port_attr.lid,
-            gid: self.ctx.gid,
+            gid: Some(self.ctx.gid),
         }
     }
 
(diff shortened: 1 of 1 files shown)


ibverbs/src/lib.rs
```diff
@@ -953,7 +958,6 @@ impl<'res> PreparedQueuePair<'res> {
     /// max_rd_atomic = 1;
     ///
     /// ah_attr.sl = 0;
-    /// ah_attr.is_global = 1;
     /// ah_attr.src_path_bits = 0;
     /// ah_attr.grh.hop_limit = 0xff;
```
jonhoo, line 962: I think this should then also be removed?

ibverbs/src/lib.rs
```diff
@@ -938,6 +941,8 @@ impl<'res> PreparedQueuePair<'res> {
     /// Internally, this uses `ibv_modify_qp` to mark the `QueuePair` as initialized
     /// (`IBV_QPS_INIT`), ready to receive (`IBV_QPS_RTR`), and ready to send (`IBV_QPS_RTS`).
     /// Further discussion of the protocol can be found on [RDMAmojo].
+    /// 
+    /// If the endpoint contains a Gid, the routing will be global.
```
jonhoo, line 945: Let's also write out what the implication is of this in terms of what fields will be set on `ah_attr`.

jonhoo (changes requested):
Seems reasonable to me! Just a few smaller bits:

jonhoo:
Looks like rustfmt also [isn't happy](https://github.com/jonhoo/rust-ibverbs/actions/runs/7593405746/job/20683850564?pr=26#step:4:22) with this line:

https://github.com/jonhoo/rust-ibverbs/blob/53e1782306ab67ff786a0ac9391fffd070c47702/ibverbs/src/lib.rs#L1410

YtvwlD:
Good points. I've modified the documentation and the line in the test.

jonhoo (approved):
Looks good to me, thanks!

jonhoo:
Released as 0.7.1 :tada:

PR #30, opened 2024-02-06
 Set attributes only on queue pairs of the correct type 

YtvwlD:
Some of the Queue Pair attributes are only valid for connections and some are only valid for reliable connections.
Setting an attribute that is not valid, might produce `EINVAL` (at least, it did for me. I'm not sure from the man pages whether this always happens).

So, what I've done is add setters for those fields that did not already have one and then, convert those fields to `Option`s and add `panic`s to those setters.

I'm not quite satisfied with panicking and I already thought about replacing the fields (and the type field) with two optional structs named `Connection` and `Reliability` (?), but then I don't know how to set them. Plus, the approach I took is compatible with existing code.

I only modified the fields on `PreparedQueuePair` and `QueuePairBuilder`. This is enough to get a UC QP up and running, for UD, the `Endpoint` struct would probably need to change.

### DIFF
--- a/ibverbs/src/lib.rs
+++ b/ibverbs/src/lib.rs
@@ -549,15 +549,24 @@ pub struct QueuePairBuilder<'res> {
     qp_type: ffi::ibv_qp_type::Type,
 
     // carried along to handshake phase
-    access: ffi::ibv_access_flags,
-    timeout: u8,
-    retry_count: u8,
-    rnr_retry: u8,
-    min_rnr_timer: u8,
-    max_rd_atomic: u8,
-    max_dest_rd_atomic: u8,
-    path_mtu: u32,
-    rq_psn: u32,
+    /// only valid for RC and UC
+    access: Option<ffi::ibv_access_flags>,
+    /// only valid for RC
+    timeout: Option<u8>,
+    /// only valid for RC
+    retry_count: Option<u8>,
+    /// only valid for RC
+    rnr_retry: Option<u8>,
+    /// only valid for RC
+    min_rnr_timer: Option<u8>,
+    /// only valid for RC
+    max_rd_atomic: Option<u8>,
+    /// only valid for RC
+    max_dest_rd_atomic: Option<u8>,
+    /// only valid for RC and UC
+    path_mtu: Option<u32>,
+    /// only valid for RC and UC
+    rq_psn: Option<u32>,
 }
 
 impl<'res> QueuePairBuilder<'res> {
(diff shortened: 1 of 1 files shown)


jonhoo:
Good catch, thanks for reporting back!

I actually wonder if we want to go a slightly different route here: don't panic in the setter, but instead document that these options "do nothing" on inappropriate types, and then keep the `Option` + only set if `Some` behavior you have in here. I think that's the path that's least likely to break existing users.

For the next major version, what I would probably do is parameterize the `PreparedQueuePair` and `QueuePairBuilder` by a generic parameter `QPT`, which can be either `struct Connection` or `struct ReliableConnection`. You could then put all the settings that are only valid for particular types of queue pairs inside those structs, and require that they be configured that way. Doing it like this would ensure that no invalid options are set at compile time, rather than relying on runtime checks and panics or `Result`s.

Which raises the question: would you be interested in picking that effort up? I'm not actually using this crate myself any more, so am arguably not the best custodian for it. I'd be happy to give you commit access to the repo, and eventually publish rights to crates.io!

YtvwlD:
Hm, I think just doing nothing in the setters would be confusing. But I would be fine with `warn!` instead of `panic!`.

I really like the idea of ensuring the correctness at compile time, though I don't think distinguishing between just `Connection` and `ReliableConnection` would be a good way forward. There's quite a bit of overlap there and still, there would be no support for `UD` queue pairs. I was thinking more about having a field `PreparedQueuePair` and `QueuePairBuilder` in the form of a `Option<Connection>`:
If it's not there, it's `UD`. And this struct could have an additional `Option<Reliability>` which would differentiate between `UC` and `RC`.
I think this would be a relatively exact mapping, but I'm not sure how ergonomic these nested structs would be. There would probably need to be builders for each of them.

With regards to the commit rights: I'm just taking my first steps with RDMA and `ibverbs` at the moment (and the documentation of this crate really helped, thank you!), so I'm not sure I would be the right fit (though I'm hoping to get better at this in the next weeks). I have [three very small crates](https://crates.io/users/YtvwlD) at crates.io, so I know the basic stuff, but just that.

So, for the next steps, I'm going to replace the `panic`s I've added with `warn`s. :)

jonhoo:
Ah, so, to be clear, I wasn't proposing separate types for them, I was proposing a single generic type with type _parameters_ for things like reliability. So, for example, `PreparedQueuePair<Connection, Reliable>` or `PreparedQueuePair<Unidirectional, Unreliable>`.

I'd like to avoid taking a dependency on `log` (or a logging crate) just for this. Besides, those logs will still almost certainly be missed by consumers anyway since nothing will happen if they haven't set up logging! So let's leave them out and go with the "silently ignored" (+ documentation!) for now :)

YtvwlD:
> Ah, so, to be clear, I wasn't proposing separate types for them, I was proposing a single generic type with type parameters for things like reliability. So, for example, `PreparedQueuePair<Connection, Reliable>` or `PreparedQueuePair<Unidirectional, Unreliable>`.

Ah yes, that might work. I might have to think a bit about this. :)

> I'd like to avoid taking a dependency on log (or a logging crate) just for this. Besides, those logs will still almost certainly be missed by consumers anyway since nothing will happen if they haven't set up logging! So let's leave them out and go with the "silently ignored" (+ documentation!) for now :)

You're probably right about people not seeing warnings. :(

I'm still not entirely happy with function calls that do nothing, but then again, it's better than erroring with a non-descript `EINVAL`.

YtvwlD:
This is a no-change rebase. Please ping me if I should change something. :)

ibverbs/src/lib.rs
```diff
@@ -597,34 +610,58 @@ impl<'res> QueuePairBuilder<'res> {
...
+    /// Valid only for RC and UC QPs.
     pub fn allow_remote_rw(&mut self) -> &mut Self {
-        self.access = self.access
-            | ffi::ibv_access_flags::IBV_ACCESS_REMOTE_WRITE
-            | ffi::ibv_access_flags::IBV_ACCESS_REMOTE_READ;
+        if self.qp_type == ffi::ibv_qp_type::IBV_QPT_RC
+            || self.qp_type == ffi::ibv_qp_type::IBV_QPT_UC
+        {
+            self.access = Some(
+                self.access.unwrap()
```
jonhoo, line 653:
```suggestion
                self.access.expect("always set to Some in new")
```

ibverbs/src/lib.rs
```diff
@@ -736,13 +780,68 @@ impl<'res> QueuePairBuilder<'res> {
...
+    pub fn set_max_rd_atomic(&mut self, max_rd_atomic: u8) -> &mut Self {
+        if self.qp_type == ffi::ibv_qp_type::IBV_QPT_RC {
+            self.max_rd_atomic = Some(max_rd_atomic);
+        }
+        self
+    }
+
+    /// Set the number of responder resources for handling incoming RDMA reads & atomic operations.
+    ///
+    /// This defaults to 1.
```
jonhoo, line 809:
```suggestion
    /// This defaults to 1.
    /// Valid only for RC QPs.
```

ibverbs/src/lib.rs
```diff
@@ -736,13 +780,68 @@ impl<'res> QueuePairBuilder<'res> {
...
+    pub fn set_max_dest_rd_atomic(&mut self, max_dest_rd_atomic: u8) -> &mut Self {
+        if self.qp_type == ffi::ibv_qp_type::IBV_QPT_RC {
+            self.max_dest_rd_atomic = Some(max_dest_rd_atomic);
+        }
+        self
+    }
+
+    /// Set the path MTU.
+    ///
+    /// Defaults to the port's active_mtu.
```
jonhoo, line 819:
```suggestion
    /// Defaults to the port's active_mtu.
    /// Valid only for RC and UC QPs.
```

ibverbs/src/lib.rs
```diff
@@ -736,13 +780,68 @@ impl<'res> QueuePairBuilder<'res> {
...
+        {
+            assert!((1..=5).contains(&path_mtu));
+            self.path_mtu = Some(path_mtu);
+        }
+        self
+    }
+
+    /// Set the PSN for the receive queue.
+    ///
+    /// Defaults to 0.
```
jonhoo, line 838:
```suggestion
    /// Defaults to 0.
    /// Valid only for RC and UC QPs.
```

jonhoo (approved):
Looks good to me with some minor nits!

I see you left two TODOs — should they block merging?

YtvwlD:
Ah, yes, good catch. :)

I'm personally fine with leaving the TODOs for the moment, because I currently just need UC and RC. :upside_down_face: 
I also don't really know much about UD, so making this work would be more difficult.

jonhoo:
Then let's get this out! Released as 0.8.1 :tada: 
Thanks :D

Issue #31, opened 2024-02-21
Modify Memory Region Protection


cholcombe973:
It mentions in the docs here that the memory region allocation from a protection domain has a default set of access rights: 
https://github.com/jonhoo/rust-ibverbs/blob/main/ibverbs/src/lib.rs#L1173
Is it possible to modify these rights? I'm new to this crate so I apologize if I overlooked something.

PR #32, opened 2024-02-21
Accept access flags for allocate

cholcombe973:
Addresses #31. Modifies accept to take an optional set of access flags. Pretty straight forward :)

### DIFF
--- a/ibverbs/src/lib.rs
+++ b/ibverbs/src/lib.rs
@@ -1144,7 +1144,7 @@ impl<'ctx> ProtectionDomain<'ctx> {
     /// `device_attr.max_mr_size`. There isn't any way to know what is the total size of memory
     /// that can be registered for a specific device.
     ///
-    /// `allocate` currently sets the following permissions for each new `MemoryRegion`:
+    /// `allocate` accepts an optional set of permission flags. The default is:
     ///
     ///  - `IBV_ACCESS_LOCAL_WRITE`: Enables Local Write Access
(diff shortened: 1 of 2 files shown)


jonhoo:
Thanks for the PR! Let's add this as a new method `alloc_with_permissions` (or something like it), and then have the existing `allocate` call that one with the default flags. That way the `Option` isn't needed and we avoid this needing to be breaking change to the API :)

cholcombe973:
Sounds good! I'll have an update coming in a few mins

cholcombe973:
Let me know if you'd like me to adjust the documentation. I feel like maybe it's a good idea to point out what each permission flag does.

ibverbs/src/lib.rs
```diff
@@ -1197,6 +1190,48 @@ impl<'ctx> ProtectionDomain<'ctx> {
...
+    ///
+    /// Every successful registration will result with a MR which has unique (within a specific
+    /// RDMA device) `lkey` and `rkey` values. These keys must be communicated to the other end's
+    /// `QueuePair` for direct memory access.
+    ///
+    /// The maximum size of the block that can be registered is limited to
+    /// `device_attr.max_mr_size`. There isn't any way to know what is the total size of memory
+    /// that can be registered for a specific device.
+    ///
+    /// `allocate` currently sets the following permissions for each new `MemoryRegion`:
```
jonhoo, line 1209: Let's also add a line after the list that point at `allocate_with_permissions` if users need more control :+1:

jonhoo (approved):
Thanks!

PR #36, opened 2024-07-20
Build outside of source dir (using cmake crate)

jonhoo:
Locally I also need https://github.com/linux-rdma/rdma-core/pull/1485 for this to work.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -50,7 +50,7 @@ jobs:
         with:
           submodules: true
       - name: Configure CI host
-        run: configure-ci.sh
+        run: ./configure-ci.sh
       - name: Install ${{ matrix.toolchain }}
         uses: dtolnay/rust-toolchain@master
         with:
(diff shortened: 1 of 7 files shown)


jonhoo:
_sigh_ as far as I can tell this is failing because the path we're building in is now too long to fit inside of `sun_path` (the path part of UNIX domains sockets ..

jonhoo:
Ref <https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars>

jonhoo:
Fixed by making the target directory be at `/tmp/cargo` instead.

The move to Clang is probably actually unnecessary now, but :shrug:

jonhoo:
Also worth noting that this change means we _also_ run `cmake --build` (not just `cmake`), which I think is technically unnecessary to get the input we need for `bindgen`. Hoping that won't cause major headache for folks. There unfortunately doesn't appear to be a way to [make the `cmake` crate](https://docs.rs/cmake/0.1.50/cmake/struct.Config.html) not call `--build`.

jonhoo:
Filed a ticket asking for this functionality in the `cmake` crate: https://github.com/rust-lang/cmake-rs/issues/207

jonhoo:
Also, FYI, after I change the invocation to not also call the `install` target I no longer need https://github.com/linux-rdma/rdma-core/pull/1485 (since we never hit that step).

jonhoo:
Released in `ibverbs-sys 0.2.1` :tada:

PR #38, opened 2024-10-30
Make work queue limits configurable in QueuePairBuilder.

sensille:
In create_qp, max_send_wr and max_recv_wr are set to a fixed value of 1. Does this have a specific reason? I would like to use larger values in my application.


jonhoo:
Hmm, it's been a very long time since I used this myself, so I honestly don't remember. Seems fine to allow configuring them though, thanks!

jonhoo (approved):

PR #47, opened 2025-03-08
cq wait using fd

xmakro:
Waits for one or more work completions in a Completion Queue (CQ). This is useful to avoid spin-looping in low-traffic scenarios

### DIFF
--- a/ibverbs/Cargo.toml
+++ b/ibverbs/Cargo.toml
@@ -20,6 +20,7 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 ffi = { path = "../ibverbs-sys", package = "ibverbs-sys", version = "0.3.0" }
+libc = "0.2"
 
 [dependencies.serde]
 version = "1.0.100"
--- a/ibverbs/Cargo.toml
+++ b/ibverbs/Cargo.toml
@@ -20,10 +20,10 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 ffi = { path = "../ibverbs-sys", package = "ibverbs-sys", version = "0.3.0" }
-nix = { version = "0.29", default-features = false, features = ["fs", "poll"] }
+nix = { version = "0.29.0", default-features = false, features = ["fs", "poll"] }
 
 [dependencies.serde]
-version = "1"
+version = "1.0.100"
 optional = true
 features = ["derive"]
 
--- a/ibverbs/Cargo.toml
+++ b/ibverbs/Cargo.toml
@@ -20,10 +20,10 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 ffi = { path = "../ibverbs-sys", package = "ibverbs-sys", version = "0.3.0" }
-libc = "0.2"
+nix = { version = "0.29", default-features = false, features = ["fs", "poll"] }
 
 [dependencies.serde]
-version = "1.0.100"
+version = "1"
 optional = true
 features = ["derive"]
 
--- a/ibverbs/src/lib.rs
+++ b/ibverbs/src/lib.rs
@@ -644,7 +644,7 @@ impl<'ctx> CompletionQueue<'ctx> {
             }
 
             assert_eq!(self.cq, out_cq);
-            // cq_context is the opaque user defined identifier passed to `ibv_create_cq()``.
+            // cq_context is the opaque user defined identifier passed to `ibv_create_cq()`.
             assert!(out_cq_context.is_null());
 
             // All completion events returned by ibv_get_cq_event() must eventually be acknowledged with ibv_ack_cq_events().
--- a/ibverbs/src/lib.rs
+++ b/ibverbs/src/lib.rs
@@ -72,6 +72,7 @@ use std::marker::PhantomData;
 use std::mem;
 use std::os::raw::c_void;
 use std::ptr;
+use std::time::Duration;
 
 const PORT_NUM: u8 = 1;
 
@@ -424,12 +425,28 @@ impl Context {
     ///  - `EINVAL`: Invalid `min_cq_entries` (must be `1 <= cqe <= dev_cap.max_cqe`).
     ///  - `ENOMEM`: Not enough resources to complete this operation.
     pub fn create_cq(&self, min_cq_entries: i32, id: isize) -> io::Result<CompletionQueue<'_>> {
+        let cc = unsafe { ffi::ibv_create_comp_channel(self.ctx) };
+        if cc.is_null() {
+            return Err(io::Error::last_os_error());
+        }
+
+        let cc_fd = unsafe { *cc }.fd;
+        let flags = unsafe { libc::fcntl(cc_fd, libc::F_GETFL) };
+        if flags < 0 {
+            return Err(io::Error::last_os_error());
+        }
+
+        let rc = unsafe { libc::fcntl(cc_fd, libc::F_SETFL, flags | libc::O_NONBLOCK) };
+        if rc < 0 {
+            return Err(io::Error::last_os_error());
+        }
+
         let cq = unsafe {
             ffi::ibv_create_cq(
                 self.ctx,
                 min_cq_entries,
                 ptr::null::<c_void>().offset(id) as *mut _,
-                ptr::null::<c_void>() as *mut _,
+                cc,
                 0,
             )
         };
@@ -439,6 +456,7 @@ impl Context {
         } else {
             Ok(CompletionQueue {
                 _phantom: PhantomData,
+                cc,
                 cq,
             })
         }
@@ -479,6 +497,7 @@ impl Drop for Context {
 /// A completion queue that allows subscribing to the completion of queued sends and receives.
 pub struct CompletionQueue<'ctx> {
     _phantom: PhantomData<&'ctx ()>,
+    cc: *mut ffi::ibv_comp_channel,
     cq: *mut ffi::ibv_cq,
 }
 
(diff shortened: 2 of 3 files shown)


ibverbs/src/lib.rs
```diff
@@ -533,6 +552,94 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+            let ctx: *mut ffi::ibv_context = unsafe { *self.cq }.context;
+            let errno = unsafe {
+                let ops = &mut { &mut *ctx }.ops;
+                ops.req_notify_cq.as_mut().unwrap()(self.cq, 0)
+            };
+            if errno != 0 {
+                return Err(io::Error::from_raw_os_error(errno));
+            }
+
+            let completions = self.poll(unsafe { &mut *c })?;
```
jonhoo, line 597: Is another call to `poll` actually warranted here? As in, does the `req_notify_cq` make it likely that there may no be available completions even though there weren't before the call?

ibverbs/src/lib.rs
```diff
@@ -533,6 +552,94 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+            if !completions.is_empty() {
+                return Ok(completions);
+            }
+
+            let mut pollfd = libc::pollfd {
```
jonhoo, line 602: I think you could get rid of a lot of the unsafety around polling by using [`nix::poll`](https://docs.rs/nix/latest/nix/poll/index.html) instead. If you do take the dependency on `nix` instead of `libc`, would recommend doing so with `default-features = false` and _just_ enable the `poll` feature.

ibverbs/src/lib.rs
```diff
@@ -533,6 +552,94 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+            let rc = unsafe { ffi::ibv_get_cq_event(self.cc, &mut out_cq, &mut out_cq_context) };
+            if rc < 0 {
+                let e = io::Error::last_os_error();
+                if e.kind() == io::ErrorKind::WouldBlock {
+                    continue;
+                }
+                return Err(e);
+            }
+
+            assert_eq!(self.cq, out_cq);
```
jonhoo, line 637: Why are we guaranteed that these are the same?

ibverbs/src/lib.rs
```diff
@@ -533,6 +552,94 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+                        "Timed out during completion queue wait",
+                    ))
+                }
+                1 => {}
+                _ => unreachable!(),
+            }
+
+            let mut out_cq = std::ptr::null_mut();
+            let mut out_cq_context = std::ptr::null_mut();
+            let rc = unsafe { ffi::ibv_get_cq_event(self.cc, &mut out_cq, &mut out_cq_context) };
```
jonhoo, line 628: Should we also assert that `out_cq_context` matches our existing context?

ibverbs/src/lib.rs
```diff
@@ -533,6 +552,94 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+                let e = io::Error::last_os_error();
+                if e.kind() == io::ErrorKind::WouldBlock {
+                    continue;
+                }
+                return Err(e);
+            }
+
+            assert_eq!(self.cq, out_cq);
+            unsafe {
+                ffi::ibv_ack_cq_events(self.cq, 1);
```
jonhoo, line 639: The 1 here looks wrong — what if there were multiple events? Should this use `rc`? If not, is `rc` ever `>0`, and if so, what does that mean?

ibverbs/src/lib.rs
```diff
@@ -533,6 +552,94 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+            }
+
+            let mut out_cq = std::ptr::null_mut();
+            let mut out_cq_context = std::ptr::null_mut();
+            let rc = unsafe { ffi::ibv_get_cq_event(self.cc, &mut out_cq, &mut out_cq_context) };
+            if rc < 0 {
+                let e = io::Error::last_os_error();
+                if e.kind() == io::ErrorKind::WouldBlock {
+                    continue;
+                }
```
jonhoo, line 633: If `ibv_get_cq_event` supports blocking operations, why do we even need `libc::poll`? Couldn't we just use `ibv_get_cq_event` directly with a blocking socket? Is the only thing preventing us from doing that the lack of a timeout? If so, would be worthwhile to add a comment to that effect.

jonhoo (changes requested):

ibverbs/Cargo.toml
```diff
@@ -20,10 +20,10 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 ffi = { path = "../ibverbs-sys", package = "ibverbs-sys", version = "0.3.0" }
-libc = "0.2"
+nix = { version = "0.29", default-features = false, features = ["fs", "poll"] }
 
 [dependencies.serde]
-version = "1.0.100"
+version = "1"
```
jonhoo, line 26: You'll generally want to keep minor+patch versions in dependencies so that dependents don't accidentally end up with older versions which do not compile with our crate.

ibverbs/src/lib.rs
```diff
@@ -533,6 +553,97 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+    /// - `TimedOut`: If the timeout expires before any completions are available.
+    /// - System errors: From underlying calls like `req_notify_cq`, `poll`, or `ibv_get_cq_event`.
+    pub fn wait<'c>(
+        &self,
+        completions: &'c mut [ffi::ibv_wc],
+        timeout: Option<Duration>,
+    ) -> io::Result<&'c mut [ffi::ibv_wc]> {
+        let c = completions as *mut [ffi::ibv_wc];
+
+        //
```
jonhoo, line 577: nit: leftover comment?

ibverbs/src/lib.rs
```diff
@@ -533,6 +553,97 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+        loop {
+            let polled_completions = self.poll(unsafe { &mut *c })?;
+            if !polled_completions.is_empty() {
+                return Ok(polled_completions);
+            }
+
+            // SAFETY: dereferencing completion queue, which is guaranteed to not have been destroyed yet.
```
jonhoo, line 584: > because we don't destroy it until in `Drop`, and given we have `self`, `Drop` has not been called

ibverbs/src/lib.rs
```diff
@@ -533,6 +553,97 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+    pub fn wait<'c>(
+        &self,
+        completions: &'c mut [ffi::ibv_wc],
+        timeout: Option<Duration>,
+    ) -> io::Result<&'c mut [ffi::ibv_wc]> {
+        let c = completions as *mut [ffi::ibv_wc];
+
+        //
+        loop {
+            let polled_completions = self.poll(unsafe { &mut *c })?;
```
jonhoo, line 579: Instead of having the `unsafe` here, can't you just use `completions` directly and only cast it to `c` until below this?

ibverbs/src/lib.rs
jonhoo: And I assume this applies transitively to the context?

ibverbs/src/lib.rs
```diff
@@ -533,6 +553,97 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+            let errno = unsafe {
+                let ops = &mut { &mut *ctx }.ops;
+                ops.req_notify_cq.as_mut().unwrap()(self.cq, 0)
+            };
+            if errno != 0 {
+                return Err(io::Error::from_raw_os_error(errno));
+            }
+
+            // We poll again to avoid a race when Work Completions arrive between the first `poll()` and `req_notify_cq()`.
+            let polled_completions = self.poll(unsafe { &mut *c })?;
```
jonhoo, line 595: Same here — I think this can still use `completions` rather than `c`?

ibverbs/src/lib.rs
```diff
@@ -424,12 +426,24 @@ impl Context {
...
     pub fn create_cq(&self, min_cq_entries: i32, id: isize) -> io::Result<CompletionQueue<'_>> {
+        let cc = unsafe { ffi::ibv_create_comp_channel(self.ctx) };
+        if cc.is_null() {
+            return Err(io::Error::last_os_error());
+        }
+
+        let cc_fd = unsafe { *cc }.fd;
+        let flags = nix::fcntl::fcntl(cc_fd, nix::fcntl::F_GETFL)?;
+        let arg = nix::fcntl::FcntlArg::F_SETFL(
+            nix::fcntl::OFlag::from_bits_retain(flags) | nix::fcntl::OFlag::O_NONBLOCK,
```
jonhoo, line 437: It's surprising to me that this is even necessary — I thought `poll` allowed you to poll with a timeout even for blocking operations?

ibverbs/src/lib.rs
```diff
@@ -533,6 +553,97 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+            )?;
+            match ret {
+                0 => {
+                    return Err(io::Error::new(
+                        io::ErrorKind::TimedOut,
+                        "Timed out during completion queue wait",
+                    ))
+                }
+                1 => {}
+                _ => unreachable!(),
```
jonhoo, line 621:
```suggestion
                _ => unreachable!("we passed 1 fd to poll, but it returned {ret}"),
```
so that we get something useful out if this ever _does_ trigger

ibverbs/src/lib.rs
```diff
@@ -533,6 +553,97 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+            }
+
+            let mut out_cq = std::ptr::null_mut();
+            // The cq_context is an opaque identifier that
```
jonhoo, line 625: Incomplete comment?

ibverbs/src/lib.rs
```diff
@@ -533,6 +553,97 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+            let mut out_cq_context = std::ptr::null_mut();
+            // The Completion Notification must be read using ibv_get_cq_event().
+            // SAFETY: c ffi call
```
jonhoo, line 628: I'm excited for https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-extern.html

ibverbs/src/lib.rs
jonhoo: Ah yes, this is https://github.com/rust-lang/rust/issues/54663 :disappointed: Worth adding that to a comment for why this needs to be `unsafe` then.

ibverbs/src/lib.rs
```diff
@@ -533,6 +555,103 @@ impl<'ctx> CompletionQueue<'ctx> {
...
+            if rc < 0 {
+                let e = io::Error::last_os_error();
+                if e.kind() == io::ErrorKind::WouldBlock {
+                    continue;
+                }
+                return Err(e);
+            }
+
+            assert_eq!(self.cq, out_cq);
+            // cq_context is the opaque user defined identifier passed to `ibv_create_cq()``.
```
jonhoo, line 647:
```suggestion
            // cq_context is the opaque user defined identifier passed to `ibv_create_cq()`.
```

ibverbs/src/lib.rs
jonhoo: But we call `ibv_get_cq_event` after `poll` returns telling us there is something to read, so it shouldn't block? Not a blocking (heh) concern, I'm just surprised it's needed.

jonhoo (approved):
Excellent, thank you!

PR #49, opened 2025-03-08
requery git before creating qp

xmakro:
I found that the git table can change over the lifetime of an ibvcontext, so the port attrs need to be requeried. this pr fixes that. 

it is a breaking change, so a new major version needs to be released. maybe we can bundle a few more changes before releasing the new major version.

### DIFF
--- a/ibverbs/src/lib.rs
+++ b/ibverbs/src/lib.rs
@@ -74,7 +74,6 @@ use std::os::raw::c_void;
 use std::ptr;
 
 const PORT_NUM: u8 = 1;
-const MAX_GID_TBL_LEN: usize = 32;
 
 /// Direct access to low-level libverbs FFI.
 pub use ffi::ibv_gid_type;
@@ -381,6 +380,7 @@ impl Context {
         }
 
         let ctx = Context { ctx };
+        // checks that the port is active/armed.
         ctx.query_port()?;
         Ok(ctx)
     }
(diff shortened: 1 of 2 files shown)


ibverbs/src/lib.rs
```diff
@@ -74,6 +74,7 @@ use std::os::raw::c_void;
 use std::ptr;
 
 const PORT_NUM: u8 = 1;
+const MAX_GID_TBL_LEN: usize = 32;
```
jonhoo, line 77: Is this const defined in such a way that we can get it out of `ibverbs-sys` instead somehow?

ibverbs/src/lib.rs
```diff
@@ -377,31 +365,24 @@ impl Context {
...
-        gid_table.truncate(num_entries as usize);
-        let gid_table = gid_table.into_iter().map(GidEntry::from).collect();
 
-        Ok(Context {
-            ctx,
-            port_attr,
-            gid_table,
-        })
+        let ctx = Context { ctx };
+        ctx.query_port()?;
```
jonhoo, line 384: So we don't need to store the result of this call any more because it actually needs to be queried per QP, not just once for the whole device context? In that case, why even have the call here at all?

ibverbs/src/lib.rs
```diff
@@ -464,8 +445,20 @@ impl Context {
     }
 
     /// Returns the valid GID table entries of this RDMA device context.
-    pub fn gid_table(&self) -> &[GidEntry] {
-        &self.gid_table
+    pub fn gid_table(&self) -> io::Result<Vec<GidEntry>> {
+        let mut gid_table = vec![ffi::ibv_gid_entry::default(); MAX_GID_TBL_LEN];
+        let num_entries = unsafe {
```
jonhoo, line 450: This needs to check for whether `num_entries < 0` and if so error still I think.

ibverbs/src/lib.rs
```diff
@@ -1127,19 +1124,24 @@ impl<'res> PreparedQueuePair<'res> {
...
-            // NOTE: bounds check happened in `set_gid_index`.
-            let gid_entry = &self.ctx.gid_table[gid_index as usize];
-            assert_eq!(gid_entry.gid_index, gid_index);
-            gid_entry.gid
-        });
-        QueuePairEndpoint {
+        let gid = if let Some(gid_index) = self.gid_index {
+            let mut gid = ffi::ibv_gid::default();
+            let rc =
+                unsafe { ffi::ibv_query_gid(self.ctx.ctx, PORT_NUM, gid_index as i32, &mut gid) };
```
jonhoo, line 1132: I'm surprised we no longer need to look in the `gid_table` any more now —does `ibv_query_gid` suffice?

jonhoo (changes requested):
Thanks for catching this!

Given I end up having very little time for this crate these days, and also don't have an actual use for it (and so my knowledge of the use-cases is limited), I'd love to have you on as a co-maintainer! I've given you permissions on the repo now so that you don't necessarily need to block on me (though I'm happy to continue reviewing!). Down the line we can also chat about getting you permissions to publish new versions to crates.io :)

jonhoo (approved):

jonhoo:
Amazing, thank you! Please then do feel free to start taking a look at the other open issues/PRs and see if you can help get some of them over the line!

PR #50, opened 2025-03-08
allow sending/receiving multiple sge slices

xmakro:


ibverbs/src/lib.rs
```diff
@@ -1304,6 +1301,51 @@ impl<T> MemoryRegion<T> {
...
+    }
+    .checked_mul(size_of::<T>())
+    .unwrap();
+    let end = match bounds.end_bound() {
+        std::ops::Bound::Included(i) => (*i + 1).checked_mul(size_of::<T>()).unwrap(),
+        std::ops::Bound::Excluded(i) => i.checked_mul(size_of::<T>()).unwrap(),
+        std::ops::Bound::Unbounded => bytes_len,
+    };
+    assert!(start <= bytes_len);
+    assert!(end <= bytes_len);
```
jonhoo, line 1338: Should we also assert that `start < end`?

ibverbs/src/lib.rs
```diff
@@ -1488,26 +1530,16 @@ impl<'res> QueuePair<'res> {
     ///
     /// [1]: http://www.rdmamojo.com/2013/01/26/ibv_post_send/
     #[inline]
-    pub unsafe fn post_send<T, R>(
+    pub unsafe fn post_send<'a>(
         &mut self,
-        mr: &mut MemoryRegion<T>,
-        range: R,
+        local: &[LocalMemorySlice<'a>],
```
jonhoo, line 1535: I think this needs to be `&mut []` so that callers can't accidentally use the same local memory region to multiple calls at once.

ibverbs/src/lib.rs
```diff
@@ -1304,6 +1301,51 @@ impl<T> MemoryRegion<T> {
             key: unsafe { &*self.mr }.rkey,
         }
     }
+
+    /// Make a subslice of this memory region.
+    pub fn slice(&mut self, bounds: impl RangeBounds<usize>) -> LocalMemorySlice<'_> {
```
jonhoo, line 1306: Because this takes `&mut self`, you'll still only be able to construct one such sub-slice at a time, which feels like it negates the intended outcome? In particular, you will _only_ be able to send/receive to multiple memory regions if they are sliced from _different_ `MemoryRegion` as currently written.

ibverbs/src/lib.rs
```diff
@@ -1488,26 +1530,16 @@ impl<'res> QueuePair<'res> {
...
-    pub unsafe fn post_send<T, R>(
+    pub unsafe fn post_send<'a>(
         &mut self,
-        mr: &mut MemoryRegion<T>,
-        range: R,
+        local: &[LocalMemorySlice<'a>],
         wr_id: u64,
-    ) -> io::Result<()>
-    where
-        R: sliceindex::SliceIndex<[T], Output = [T]>,
```
jonhoo, line 1498: The reason why `SliceIndex` is nice is because it also works for things like solitary `usize`. It's a little sad to lose that for ergonomics, though I suppose it's not the end of the world if it enables a class of use-cases that were otherwise inaccessible.

ibverbs/src/lib.rs
jonhoo: I think what you'd have to do to enable multi-sub-slicing is the same thing `std` does for slices, which is to provide `split_at` and friends that return multiple guaranteed-non-overlapping sub-slices.

ibverbs/src/lib.rs
```diff
@@ -1488,26 +1530,16 @@ impl<'res> QueuePair<'res> {
...
-            length: mem::size_of_val(range) as u32,
-            lkey: (*mr.mr).lkey,
-        };
+    ) -> io::Result<()> {
         let mut wr = ffi::ibv_send_wr {
             wr_id,
             next: ptr::null::<ffi::ibv_send_wr>() as *mut _,
-            sg_list: &mut sge as *mut _,
-            num_sge: 1,
+            sg_list: local.as_ptr() as *mut ffi::ibv_sge,
```
jonhoo, line 1541: I don't think this works at the moment (or if so, it's accidentaly), because it requires that `LocalMemorySlice` has exactly the same layout as `ibv_sge`. To fix this you'd have to add `#[repr(transparent)]` to `LocalMemorySlice`.

ibverbs/src/lib.rs
```diff
@@ -1664,4 +1686,12 @@ mod test_serde {
         println!("{:#08x}", guid.oui());
         assert_eq!(guid.oui(), 0x123456);
     }
+
+    #[test]
+    fn test_local_memory_slice_sge_memory_layout() {
+        assert_eq!(
+            std::mem::size_of::<LocalMemorySlice<'static>>(),
+            std::mem::size_of::<ffi::ibv_sge>()
+        );
```
jonhoo, line 1695: This is good, but I think you also need to check the `align_of`.

jonhoo (changes requested):

ibverbs/src/lib.rs
jonhoo: Hmm, if these can be accessed after the fact, then `&` and `unsafe` isn't the fix either, since Rust still has a very fundamental assumption that things behind `&` will _never_ change while the `&` exists. This feeds into things like the optimizer, which is allowed to cache operations over values behind a `&` for example. The _only_ way it's legal in Rust to mutate through `&` is via `UnsafeCell` (or something that uses it internally).

It this suggests that we'll have to move to raw pointers (ie, `*mut`) for this API.

PR #52, opened 2025-03-10
Use Arc for ownership instead of references and lifetime annotations

### LINKED ISSUE #48: Use Arc for ownership instead of references and lifetime annotations
This proposes to use shared pointers (Arc) for managing the lifetime of QueuePair, CompletionQueue, ProtectionDomain and Context. All these objects are usually long lived and not in the critical path, so the extra overhead from creating the Arc and managing the refcount should be minimal. The Arc would be internal to the struct and not visible to the user. The objects would remain not clonable.

I have been using the ibverbs library extensively and found the lifetime annotations prohibitive. It is not possible to move a QueuePair, CompletionQueue, ProtectionDomain and Context into a struct without unsafe code, as they all depend on each other (even with ouroboros it is very painful).

@jonhoo please let me know if this sounds reasonable, then I will send a PR

xmakro:
closes #48


jonhoo:
@xmakro Sorry, this completely dropped off my radar. I suggest you start simply merging things yourself as I won't have reviewing capacity for a while. Please do ping me (email is arguably best) when you want to make a release though!

jonhoo:
Same goes for addressing the various other issues that have come up, like #58 :sweat_smile:

PR #61, opened 2025-07-06
Allow registering external memory regions

xmakro:


ibverbs/src/lib.rs
```diff
@@ -1661,42 +1644,14 @@ impl ProtectionDomain {
     ///  - `EINVAL`: Invalid access value.
     ///  - `ENOMEM`: Not enough resources (either in operating system or in RDMA device) to
     ///    complete this operation.
-    pub fn allocate_with_permissions<T: Sized + Copy + Default>(
+    pub fn allocate_with_permissions(
         &self,
         n: usize,
         access_flags: ffi::ibv_access_flags,
-    ) -> io::Result<MemoryRegion<T>> {
+    ) -> io::Result<MemoryRegion<Vec<u8>>> {
```
BGR360, line 1651: It is not sound for this to return a `MemoryRegion<Vec<_>>`. The caller could call `inner()` and then resize the vec, which completely reallocates the data in a region of memory not managed by libibverbs

PR #72, opened 2025-08-01
Allow specifying send flags.

e-oak:
Creates versions of Send/Read/Write with an `options` argument. For now that field only controls send flags, we could use to add additional arguments (e.g. qp_type) to those methods later on without breaking existing code if needed.


jeronimosg:
Having an iterator would be nice for ergonomics, and it works well for now. 
Eventually though, we will need to look into adding `next` to issue multiple WRs in a single `ibv_post_send`, ideally without extra allocations or copies into C land

xmakro:
Superseeded by https://github.com/jonhoo/rust-ibverbs/pull/92

PR #79, opened 2025-10-14
Add convenient struct wrappers for buffer pooling

arajni3:
Add OwnedMemoryRegion and separate MemorySlicer to allow slicing while MemoryRegion is already owned.


xmakro:
Resolved by https://github.com/jonhoo/rust-ibverbs/pull/80 instead

PR #81, opened 2025-11-05
Add EFA support to rust-ibverbs with working example.

e-ngo:
# Adding EFA Support to `rust-ibverbs`

**Tested on:** AWS instances with EFA NICs

## Key Parity Functions Added

- **`build_efa()`**  
  Mirrors `build()`, but uses the extended `libibverbs` API to create an **SRD QP**.

- **`handshake_efa()`**  
  Unlike RC QP handshakes, EFA is connectionless. This function:  
  - Transitions QP state: `INIT → RTR → RTS`  
  - Creates and binds the required **Address Handle (AH)** to authorize sends to the remote peer.

- **`post_write_efa()`**  
  Uses extended API to build a work request.  
  `ud_data.addr` must point to the target **AH**.

These functions are all hidden behind a prop passed down from the QueuePairBuilder (QueuePairBuilder.enable_efa(true)). By enabling this, the proper efa version of the method will be called.

This keeps the user's interfaces nearly the exact same.

### DIFF
--- a/ibverbs-sys/build.rs
+++ b/ibverbs-sys/build.rs
@@ -8,9 +8,9 @@ fn main() {
     println!("cargo:rustc-link-search=native={manifest_dir}/vendor/rdma-core/build/lib");
     println!("cargo:rustc-link-lib=ibverbs");
 
-    // Link EFA library when the feature is enabled
-    if env::var("CARGO_FEATURE_EFA").is_ok() {
+    if env::var("CARGO_FEATURE_EFA").is_ok() && Path::new("/opt/amazon/efa/lib/").exists() {
         println!("cargo:rustc-link-lib=efa");
+        println!("cargo:rustc-cfg=feature=\"efa\"");
     }
 
     if Path::new("vendor/rdma-core/CMakeLists.txt").exists() {
--- a/ibverbs-sys/build.rs
+++ b/ibverbs-sys/build.rs
@@ -8,11 +8,6 @@ fn main() {
     println!("cargo:rustc-link-search=native={manifest_dir}/vendor/rdma-core/build/lib");
     println!("cargo:rustc-link-lib=ibverbs");
 
-    if env::var("CARGO_FEATURE_EFA").is_ok() && Path::new("/opt/amazon/efa/lib/").exists() {
-        println!("cargo:rustc-link-lib=efa");
-        println!("cargo:rustc-cfg=feature=\"efa\"");
-    }
-
     if Path::new("vendor/rdma-core/CMakeLists.txt").exists() {
         // don't touch source dir if not necessary
     } else if Path::new(".git").is_dir() {
@@ -72,6 +67,7 @@ fn main() {
         .allowlist_function("ibv_wc_read_imm_data")
         .allowlist_function("ibv_wr_rdma_write_imm")
         .allowlist_function("ibv_wr_rdma_write")
+        .allowlist_function("ibv_wr_send")
         .allowlist_function("ibv_wr_rdma_read")
         .allowlist_type("ibv_.*")
         .allowlist_type("efa.*")
--- a/ibverbs-sys/build.rs
+++ b/ibverbs-sys/build.rs
@@ -7,8 +7,9 @@ fn main() {
     println!("cargo:include={manifest_dir}/vendor/rdma-core/build/include");
     println!("cargo:rustc-link-search=native={manifest_dir}/vendor/rdma-core/build/lib");
     println!("cargo:rustc-link-lib=ibverbs");
-    // Only link EFA if the library exists (i.e., on EFA-capable hosts)
-    if Path::new("/opt/amazon/efa/lib/").exists() {
+
+    // Link EFA library when the feature is enabled
+    if env::var("CARGO_FEATURE_EFA").is_ok() {
         println!("cargo:rustc-link-lib=efa");
     }
 
--- a/ibverbs-sys/build.rs
+++ b/ibverbs-sys/build.rs
@@ -7,6 +7,10 @@ fn main() {
     println!("cargo:include={manifest_dir}/vendor/rdma-core/build/include");
     println!("cargo:rustc-link-search=native={manifest_dir}/vendor/rdma-core/build/lib");
     println!("cargo:rustc-link-lib=ibverbs");
+    // Only link EFA if the library exists (i.e., on EFA-capable hosts)
+    if Path::new("/opt/amazon/efa/lib/").exists() {
+        println!("cargo:rustc-link-lib=efa");
+    }
 
     if Path::new("vendor/rdma-core/CMakeLists.txt").exists() {
         // don't touch source dir if not necessary
@@ -50,11 +54,28 @@ fn main() {
     eprintln!("run bindgen");
     let bindings = bindgen::Builder::default()
         .header("vendor/rdma-core/libibverbs/verbs.h")
+        .header("vendor/rdma-core/providers/efa/efadv.h")
         .clang_arg(format!("-I{built_in}/include/"))
         .allowlist_function("ibv_.*")
         .allowlist_function("_ibv_.*")
+        .allowlist_function("efadv_.*")
+        .allowlist_function("ibv_cq_ex_to_cq")
+        .allowlist_function("ibv_start_poll")
+        .allowlist_type("ibv_cq_ex")
+        .allowlist_type("ibv_poll_cq_attr")
+        .allowlist_function("ibv_next_poll")
+        .allowlist_function("ibv_end_poll")
+        .allowlist_function("ibv_wc_read_opcode")
+        .allowlist_function("ibv_wc_read_vendor_err")
+        .allowlist_function("ibv_wc_read_byte_len")
+        .allowlist_function("ibv_wc_read_imm_data")
+        .allowlist_function("ibv_wr_rdma_write_imm")
+        .allowlist_function("ibv_wr_rdma_write")
+        .allowlist_function("ibv_wr_rdma_read")
         .allowlist_type("ibv_.*")
+        .allowlist_type("efa.*")
         .allowlist_var("IBV_LINK_LAYER_.*")
+        .allowlist_var("EFADV_QP_DRIVER_TYPE_.*")
         .bitfield_enum("ibv_access_flags")
         .bitfield_enum("ibv_create_cq_wc_flags")
         .bitfield_enum("ibv_device_cap_flags")
@@ -64,6 +85,7 @@ fn main() {
         .bitfield_enum("ibv_qp_attr_mask")
         .bitfield_enum("ibv_qp_create_send_ops_flags")
         .bitfield_enum("ibv_qp_init_attr_mask")
+        .bitfield_enum("ibv_qp_init_attr_ex_mask")
         .bitfield_enum("ibv_qp_open_attr_mask")
         .bitfield_enum("ibv_raw_packet_caps")
         .bitfield_enum("ibv_rx_hash_fields")
--- a/ibverbs/src/lib.rs
+++ b/ibverbs/src/lib.rs
@@ -421,6 +421,10 @@ impl Context {
         Ok(ctx)
     }
 
+    /// Returns true if EFA is enabled for this context.
+    pub fn is_efa_enabled(&self) -> bool {
+        self.inner.enable_efa
+    }
 
     /// Create a completion queue (CQ).
     ///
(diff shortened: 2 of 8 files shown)


ibverbs/src/lib.rs
```diff
@@ -2021,6 +2160,89 @@ impl QueuePair {
...
+    /// the other side uses post_recv on a dummy buffer and get the imm data from the work completion
+    pub fn post_write_efa(
+        &mut self,
+        local: &[LocalMemorySlice],
+        remote_mr: RemoteMemorySlice,
+        remote_endpoint: QueuePairEndpoint,
+        wr_id: u64,
+        imm_data: Option<u32>,
+    ) -> io::Result<()> {
+        let qp_ex = unsafe { ffi::ibv_qp_to_qp_ex(self.qp) };
```
xmakro: check that the result is not nullptr, then dereference once and use below, then you likely dont need the unsafe after this line

ibverbs/src/lib.rs
```diff
@@ -2021,6 +2160,89 @@ impl QueuePair {
...
+    pub fn post_read_efa(
+        &mut self,
+        local: &[LocalMemorySlice],
+        remote: RemoteMemorySlice,
+        remote_endpoint: QueuePairEndpoint,
+        wr_id: u64,
+        _imm_data: Option<u32>,
+    ) -> io::Result<()> {
+
+        let qp_ex = unsafe { ffi::ibv_qp_to_qp_ex(self.qp) };
```
xmakro: same here, check this succeeded, then dereference

ibverbs/src/lib.rs
```diff
@@ -2005,12 +2121,35 @@ impl QueuePair {
...
     pub fn post_write(
         &mut self,
         local: &[LocalMemorySlice],
         remote: RemoteMemorySlice,
         wr_id: u64,
         imm_data: Option<u32>,
+    ) -> io::Result<()> {
+        if self.enable_efa {
+            // This is an EFA QP, use EFA write
+            self.post_write_efa(local, remote, self.remote_endpoint.unwrap(), wr_id, imm_data)
```
xmakro: no need to pass this because post_write_efa has access to self as well

ibverbs/src/lib.rs
```diff
@@ -1409,44 +1483,54 @@ impl PreparedQueuePair {
...
-            mask |= ffi::ibv_qp_attr_mask::IBV_QP_ACCESS_FLAGS;
+
+        if !self.enable_efa {
+            if let Some(access) = self.access {
+                attr.qp_access_flags = access.0;
+                mask |= ffi::ibv_qp_attr_mask::IBV_QP_ACCESS_FLAGS;
+            }
         }
+
+        if self.enable_efa {
```
xmakro: swap these blocks and make this
```
if self.enable_efa {
} else {...}
```

ibverbs/src/lib.rs
```diff
@@ -1391,15 +1459,21 @@ impl PreparedQueuePair {
...
     /// # Errors
     ///
     ///  - `EINVAL`: Invalid value provided in `attr` or in `attr_mask`.
     ///  - `ENOMEM`: Not enough resources to complete this operation.
     ///
     /// [RDMAmojo]: http://www.rdmamojo.com/2014/01/18/connecting-queue-pairs/
-    pub fn handshake(self, remote: QueuePairEndpoint) -> io::Result<QueuePair> {
+    pub fn handshake(mut self, remote: QueuePairEndpoint) -> io::Result<QueuePair> {
         // init and associate with port
+        let qp_type = unsafe { (*self.qp.qp).qp_type };
```
xmakro: move the qp out of self, then you dont need to make this mut self

ibverbs/src/lib.rs
```diff
@@ -1496,6 +1580,35 @@ impl PreparedQueuePair {
...
+    
+                let test_ah = unsafe { ffi::ibv_create_ah(self.qp.pd.pd, &mut ah_attr as *mut _) };
+                if test_ah.is_null() {
+                    return Err(io::Error::last_os_error());
+                }
+                test_ah
+            } else {
+                self.qp.ah
+            };
+            self.qp.ah = ah;
```
xmakro: seems a bit roundabout to store this with if else in a variable, why not just set 
`qp.ah = test_ah;` in the inf branch, and delete the else branch and let statement?

ibverbs-sys/build.rs
```diff
@@ -7,6 +7,10 @@ fn main() {
     println!("cargo:include={manifest_dir}/vendor/rdma-core/build/include");
     println!("cargo:rustc-link-search=native={manifest_dir}/vendor/rdma-core/build/lib");
     println!("cargo:rustc-link-lib=ibverbs");
+    // Only link EFA if the library exists (i.e., on EFA-capable hosts)
+    if Path::new("/opt/amazon/efa/lib/").exists() {
```
xmakro: instead of relying on a path, maybe the right thing is to feature gate this?

ibverbs/src/lib.rs
```diff
@@ -76,6 +76,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 const PORT_NUM: u8 = 1;
+const UD_QKEY: u32 = 0x12345678;
```
xmakro, line 79: what is this ?

ibverbs/src/lib.rs
```diff
@@ -1114,23 +1139,60 @@ impl QueuePairBuilder {
...
+
+        let qp = if self.enable_efa {
+            // EFA uses extended API to create an SRD QP.
+            let send_ops_flags = (ffi::ibv_qp_create_send_ops_flags::IBV_QP_EX_WITH_RDMA_WRITE.0 |
+                ffi::ibv_qp_create_send_ops_flags::IBV_QP_EX_WITH_RDMA_WRITE_WITH_IMM.0 |
+                ffi::ibv_qp_create_send_ops_flags::IBV_QP_EX_WITH_RDMA_READ.0 |
+                ffi::ibv_qp_create_send_ops_flags::IBV_QP_EX_WITH_SEND.0 |
+                ffi::ibv_qp_create_send_ops_flags::IBV_QP_EX_WITH_SEND_WITH_IMM.0) as u32;
+            let mut attr = ffi::ibv_qp_init_attr_ex {
+            qp_context: self.pd.ctx.ctx as *mut _,
```
xmakro: formatting looks broken here

ibverbs/src/lib.rs
e-ngo: QKEY is used to validate communication between two QPs, specifically for UD QPs. In this case I am using a random u32.

ibverbs/src/lib.rs
e-ngo: Gotcha. Yea i was trying to be too clever. Simplified it. Handshakes should only be called once per qp is the assumption i will make here. i think it's fair to assume?

ibverbs/src/lib.rs
e-ngo: hmm, not able to remove the full unsafe blocks., Since inner functions like `wr_rdma_write_imm` is unsafe as well.

ibverbs-sys/build.rs
e-ngo: Cool, how does this look?

ibverbs-sys/build.rs
```diff
@@ -8,6 +8,11 @@ fn main() {
     println!("cargo:rustc-link-search=native={manifest_dir}/vendor/rdma-core/build/lib");
     println!("cargo:rustc-link-lib=ibverbs");
 
+    if env::var("CARGO_FEATURE_EFA").is_ok() && Path::new("/opt/amazon/efa/lib/").exists() {
```
xmakro: since this is now feature gated, i would remove:
Path::new("/opt/amazon/efa/lib/").exists()

ibverbs/src/lib.rs
```diff
@@ -1114,23 +1142,69 @@ impl QueuePairBuilder {
...
+                    max_send_wr: self.max_send_wr,
+                    max_recv_wr: self.max_recv_wr,
+                    max_send_sge: self.max_send_sge,
+                    max_recv_sge: self.max_recv_sge,
+                    max_inline_data: self.max_inline_data,
+                },
+                qp_type: self.qp_type,
+                sq_sig_all: 0,
+            };
+            unsafe { ffi::ibv_create_qp(self.pd.pd, &mut attr as *mut _) }
```
xmakro, line 1214: I think these codepaths can be further unified by using Ibv_create_qp_ex in the non-efa case, then you can use the same ibv_qp_init_attr_ex for both cases

ibverbs-sys/build.rs
```diff
@@ -50,11 +55,28 @@ fn main() {
     eprintln!("run bindgen");
     let bindings = bindgen::Builder::default()
         .header("vendor/rdma-core/libibverbs/verbs.h")
+        .header("vendor/rdma-core/providers/efa/efadv.h")
```
xmakro, line 53: maybe the bindings for the efa headers and efa specific methods should only be created if the efa feature is enabled

ibverbs/src/lib.rs
```diff
@@ -2021,6 +2166,98 @@ impl QueuePair {
         self._post_one_sided(local, remote, wr_id, opcode, imm_data)
     }
 
+    #[inline]
+    /// Remote RDMA write with EFA.
+    /// immediate data can be used to signal the completion of the write operation
+    /// the other side uses post_recv on a dummy buffer and get the imm data from the work completion
+    pub fn post_write_efa(
```
xmakro, line 2220: I read more about the ibv ex api documentation, this is actually not EFA specific. Is using the wr-api with function calls instead of the post-api a requirement for EFA? We might be able to unify the codepaths for EFA and non EFA to both use the wr-api since it seems to be the more efficient and powerful API. Is it supported on all normal hardware these days?

ibverbs/src/lib.rs
```diff
@@ -2021,6 +2160,89 @@ impl QueuePair {
...
+            (*qp_ex).wr_flags = ffi::ibv_send_flags::IBV_SEND_SIGNALED.0;
+            if imm_data.is_some() {
+                (*qp_ex).wr_rdma_write_imm.unwrap()(qp_ex, remote_mr.rkey, remote_mr.addr, imm_data.unwrap().to_be());
+            } else {
+                (*qp_ex).wr_rdma_write.unwrap()(qp_ex, remote_mr.rkey, remote_mr.addr);
+            };
+            (*qp_ex).wr_set_sge_list.unwrap()(qp_ex, local.len(), local.as_ptr() as *mut ffi::ibv_sge);
+            // TODO(Eric): Consider creating and caching AH for each remote_endpoint...
+            // This way a single QP can be shared for multiple remote endpoints.
+            (*qp_ex).wr_set_ud_addr.unwrap()(qp_ex, self.ah, remote_endpoint.num, UD_QKEY);
```
xmakro: Since UD QueuePair works with any other UD QP, the handshake API does not make much sense for this. I'm wondering how we should change the API. Should we distinguish between ConnectedQueuePair and DatagramQueuePair?

ibverbs/src/lib.rs
e-ngo: >Is using the wr-api with function calls instead of the post-api a requirement for EFA
I believe so, primarily we need to set the ud_addr in the wr_id which seems to only be available in the extended API.

>Is it supported on all normal hardware these days?
I would assume so, the ibv_wr api seems to date back 7 years https://github.com/linux-rdma/rdma-core/blob/0d977c57c486bc13e3a6be0f17c29eace5657633/libibverbs/man/ibv_wr_post.3.md?plain=1#L29. I think that api should have backwards compatibility, as long as we try to make sure we're not setting too many other fields.

I can try to merge it. Do you mind if I do it in a follow up PR?

ibverbs/src/lib.rs
e-ngo: That's a good point. My initial reaction was that I wanted to merge the API points as much as possible so that there's less branching off in user code. What APIs would you reckon that would be different here or that we would want to distinguish?

ibverbs/src/lib.rs
sigaev: ud_addr is actually part of ibv_send_wr (it's just called "ud" there), so we can use the existing ibv_post_send API in the implementation of post_{send,write,read}:

```c
struct ibv_send_wr {
	uint64_t		wr_id;
    ...
	union {
        ...
		struct {
			struct ibv_ah  *ah;
			uint32_t	remote_qpn;
			uint32_t	remote_qkey;
		} ud;
	} wr;
```

e-ngo: I will give it a try

xmakro:
Resolved in https://github.com/jonhoo/rust-ibverbs/pull/103

PR #84, opened 2025-12-04
Allow slicing a whole memory region

arajni3:
This allows 64-bit slices unlike LocalMemoryRegion

### DIFF
--- a/ibverbs/src/lib.rs
+++ b/ibverbs/src/lib.rs
@@ -1571,6 +1571,21 @@ impl<T> MemoryRegion<T> {
             lkey: unsafe { *self.inner.mr }.lkey,
         }
     }
+
+    /// Make a subslice of this memory region.
+    pub fn slice(&self, bounds: impl RangeBounds<usize>) -> LocalMemorySlice {
+        let (addr, length) = calc_addr_len(
+            bounds,
+            unsafe { *self.inner.mr }.addr as u64,
+            unsafe { *self.inner.mr }.length,
+        );
+        let sge = ffi::ibv_sge {
+            addr,
+            length: length.try_into().unwrap(),
+            lkey: unsafe { *self.inner.mr }.lkey,
+        };
+        LocalMemorySlice { _sge: sge }
+    }
 }
 
 /// Local memory slice.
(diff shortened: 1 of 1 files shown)


ibverbs/src/lib.rs
```diff
@@ -1556,6 +1556,22 @@ impl<T> MemoryRegion<T> {
         self.data
     }
 
+    /// Make a lifetime-independent slice representation of this memory region.
+    /// Unlike `LocalMemorySlice`, this supports 64-bit memory region lengths.
+    pub fn as_slice(&self) -> MemoryRegionSlice {
+        let bounds = (..unsafe { *self.inner.mr }.length);
```
xmakro: these lines here seem obsolete, no need to recalculate the addr and length if we are taking the full range

ibverbs/src/lib.rs
```diff
@@ -1613,6 +1629,49 @@ impl LocalMemorySlice {
...
+/// Unlike `LocalMemorySlice`, this supports 64-bit memory region lengths.
+#[derive(Debug, Default, Copy, Clone)]
+pub struct MemoryRegionSlice {
+    addr: u64,
+    length: u64,
+    lkey: u32,
+}
+
+impl MemoryRegionSlice {
+    /// Get the address of the memory region.
```
xmakro: maybe just make the struct fields pub instead of these getter methods

ibverbs/src/lib.rs
```diff
@@ -1613,6 +1629,49 @@ impl LocalMemorySlice {
...
+        self.len() == 0
+    }
+
+    /// Get the local key of the memory region.
+    pub fn lkey(&self) -> u32 {
+        self.lkey
+    }
+
+    /// Make a slice of this memory region.
+    pub fn slice(&self, bounds: impl RangeBounds<usize>) -> LocalMemorySlice {
```
xmakro, line 1640: what is your use case for needing something lifetime independent?

arajni3: It can be useful to move MemoryRegion into a generic (rdma-agnostic) custom buffer pool struct, which, on drop, will call the MemoryRegion's destructor, thus deregistering the region right before deallocating the application memory. Thus, we need to be able to create slices outside of the MemoryRegion's lifetime

PR #86, opened 2026-02-28
feat: make the rdma-core paths configurable

r-vdp:
Make it possible to supply the different rdma-core paths by env vars instead of building the vendored source code.
This makes it possible to link to a system-supplied rdma-core lib.


xmakro:
Merged, thanks for the PR!

PR #92, opened 2026-06-08
Introduce WorkRequest chaining for QueuePair

aryeyur:


aryeyur:
Hey :wave: 

Would it be possible to look into this PR?

It adds support for chaining RDMA operations.

ibverbs/src/lib.rs
```diff
@@ -2246,6 +2277,183 @@ impl Drop for QueuePair {
     }
 }
 
+/// A builder to construct and post a dynamically sized chain of work requests to the Send Queue.
+pub struct ChainBuilder<'qp, 'buf> {
```
xmakro: The `ChainBuilder` only needs to know the `QueuePair` during post, so we could disentangle them until that point. This avoids the QP lifetime and allows for example creating two `ChainBuilders` at the same time, and posting them later with the same `QueuePair`. So `ChainBuilder` just becomes a helper around `&mut [ibv_send_wr]`.

If we go this route, then that raises the question if we need a `ChainBuilder` at all, or if we can have the helper only around `ibv_send_wr`, with an API looking like:

```
#[repr(transparent)]
pub struct WorkRequest<'a> {
    wr: ffi::ibv_send_wr,
    _local: PhantomData<&'a [LocalMemorySlice]>,   // keeps the sg_list borrow alive
}

impl<'a> WorkRequest<'a> {
    pub fn send (local: &'a [LocalMemorySlice], wr_id: u64, imm: Option<u32>) -> Self;
    pub fn write(local: &'a [LocalMemorySlice], remote: RemoteMemorySlice, wr_id: u64) -> Self;
    pub fn read (local: &'a [LocalMemorySlice], remote: RemoteMemorySlice, wr_id: u64) -> Self;
}
```

on the QueuePair:

```
pub unsafe fn post(&mut self, mut wrs: impl AsMut<[WorkRequest<'_>]>) -> io::Result<()>;
```

usage:

```
qp.post([
    WorkRequest::write(&locals[0], remotes[0].clone(), 1, None),
    WorkRequest::write(&locals[1], remotes[1].clone(), 2, None),
    WorkRequest::send(&notify, 3, None).signaled(),
])?;

// dynamic / reused buffer,
qp.post(&mut wrs)?;
```

Then we can also do things like:

```
WorkRequest::write(..).signaled(), .fenced()
```

What do you think? This API should have zero overhead, so we can redirect the existing single post_send/post_recv API to it.

xmakro (commented):
Thank you for the PR, please see the comment for discussion

ibverbs/src/lib.rs
aryeyur: Hey @xmakro thanks for reviewing this!

There are many upsides to this alternative, as you mentioned:
1) No lifetime coupling (this can be addressed separately).
2) Memory reuse for work requests is left to the API user in a cleaner way than requiring an "array to populate". The `poll` function does this and it's fine, but having the user handle it is even better. This also removes the need for error handling when pushing too many ops into the chain.
3) Flexibility: The API user can configure the work request however they like, e.g., enabling a signal for every operation if needed.

The one downside I see:
The chain builder "guides" the API user to include one final signaled operation. This is opinionated, but it prevents bugs that would stem from omitting it.
We can mitigate this by adding a check to ensure the last operation is signaled, which seems like a reasonable approach.
Alternatively, we could implement and document the API so that the last op is automatically signaled.
WDYT?

aryeyur: it's also nicer that it's possible to more easily reuse the work-requests construction code, instead of duplicating it in the chain-builder.
I uploaded as a separate commit on top of the 1st, I'll squash before submitting but wanted to give you a chance to compare both options before doing that.
Please LMK what you think.
Thanks!

ibverbs/src/lib.rs
```diff
@@ -2165,13 +2286,7 @@ impl QueuePair {
...
         imm_data: Option<u32>,
     ) -> io::Result<()> {
-        let opcode = if imm_data.is_some() {
-            ffi::ibv_wr_opcode::IBV_WR_RDMA_WRITE_WITH_IMM
-        } else {
-            ffi::ibv_wr_opcode::IBV_WR_RDMA_WRITE
-        };
-
-        self._post_one_sided(local, remote, wr_id, opcode, imm_data)
+        unsafe { self.post([WorkRequest::write(local, remote, wr_id, imm_data)]) }
```
xmakro: let's make `post_write` and `post_read` both `unsafe`. it is weird in the api right now that some are unsafe and some are not. also then the unsafe in the function bodies is not needed

ibverbs/src/lib.rs
```diff
@@ -2182,50 +2298,99 @@ impl QueuePair {
...
-        let anon_1 = if let Some(imm_data) = imm_data {
-            ffi::ibv_send_wr__bindgen_ty_1 {
-                imm_data: imm_data.to_be(),
-            }
-        } else {
-            Default::default()
-        };
+    /// Posts a linked list of Work Requests (WRs) to the Send Queue of this Queue Pair.
+    ///
+    /// This method automatically links the work requests together, setting the last
```
xmakro: This is the power user API that should allow maximum flexibility, so I would not set the signalled automatically. Instead, I would educate with the docs that `.signalled()` is needed for getting an event in the CQ

ibverbs/examples/doorbell_batching.rs
```diff
@@ -0,0 +1,138 @@
+use ibverbs::{LocalMemorySlice, RemoteMemorySlice, WorkRequest};
```
xmakro, line 5: Please add a brief comment at the top of the file what this example showcases

ibverbs/src/lib.rs
xmakro: Thank you, the new commit looks really great! No worries about squashing, that is automatic for this repository

xmakro:
```
---- ibverbs/src/lib.rs - QueuePair::post (line 2330) stdout ----
error[E0716]: temporary value dropped while borrowed
    --> ibverbs/src/lib.rs:2341:38
     |
2333 | fn post_many<'local>(
     |              ------ lifetime `'local` defined here
...
2341 |         wrs.push(WorkRequest::write(&[local[i]], remote[i].clone(), i as u64, None));
     |         -----------------------------^^^^^^^^^^-------------------------------------- temporary value is freed at the end of this statement
     |         |                            |
     |         |                            creates a temporary value which is freed while still in use
     |         argument requires that borrow lasts for `'local`

error: aborting due to 1 previous error
```

some docs test failure

ibverbs/src/lib.rs
aryeyur: Thanks!

ibverbs/src/lib.rs
aryeyur: done

ibverbs/src/lib.rs
aryeyur: done

ibverbs/examples/doorbell_batching.rs
aryeyur: done

ibverbs/src/lib.rs
```diff
@@ -2155,77 +2278,139 @@ impl QueuePair {
...
-        let opcode = ffi::ibv_wr_opcode::IBV_WR_RDMA_READ;
-        self._post_one_sided(local, remote, wr_id, opcode, None)
+        self.post([WorkRequest::read(local, remote, wr_id).signaled()])
     }
 
-    // internal function to do one sided communication
-    fn _post_one_sided(
+    /// Posts a linked list of Work Requests (WRs) to the Send Queue of this Queue Pair.
+    ///
+    /// This method automatically links the work requests together. Unlike the simpler helper
```
xmakro: This method modifies `wrs` to link the work requests together.

ibverbs/src/lib.rs
aryeyur: (Just making sure we're on the same page here - this is a breaking change, people who will upgrade to latest would need to add unsafe block)

ibverbs/examples/chained_ops.rs
```diff
@@ -0,0 +1,148 @@
...
+    for i in 0..num_writes {
+        wrs.push(WorkRequest::write(
+            &locals[i],
+            remotes[i].clone(),
+            (i + 1) as u64,
+            None,
+        ));
+    }
+    // Append the final Send operation to signal completion of the chain and carry the write count
+    let send_wr = WorkRequest::send(&notify_slice, send_chain_completion_wr_id, None).signaled();
```
xmakro: nit: merge these two lines

ibverbs/examples/chained_ops.rs
```diff
@@ -0,0 +1,148 @@
...
+    let dest_mr = pd.allocate(text.len()).unwrap();
+
+    let mut notify_mr = pd.allocate(NOTIFY_BUF_SIZE).unwrap();
+    let recv_mr = pd.allocate(NOTIFY_BUF_SIZE).unwrap();
+
+    // 5. Post receive request for the final send notification (4-byte payload)
+    unsafe { qp.post_receive(&[recv_mr.slice(..)], RECEIVE_NOTIFICATION_WR_ID) }.unwrap();
+
+    // 6. Split the string into slices by space delimiter and prepare the notification payload
+    let locals: Vec<[LocalMemorySlice; 1]> = text
```
xmakro: I think the `locals` and `remotes` block can be merged with:
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.unzip

ibverbs/examples/doorbell_batching.rs
```diff
@@ -0,0 +1,148 @@
+//! An example showcasing how to chain multiple Work Requests (WRs) together
```
xmakro, line 1: Please rename the file to `doorbell_batching.rs`

ibverbs/src/lib.rs
xmakro: Please correct this:

/// Posts a single send Work Request (WR) to the Send Queue of this Queue Pair.

xmakro: Please move these paragraphs to the `post` function instead:


```
    /// Generates a HW-specific Send Request for the memory at `mr[range]`, and adds it to the tail
    /// of the Queue Pair's Send Queue without performing any context switch. The RDMA device will
    /// handle it (later) in asynchronous way. If there is a failure in one of the WRs because the
    /// Send Queue is full or one of the attributes in the WR is bad, it stops immediately and
    /// return the pointer to that WR.
    ///
    /// `wr_id` is a 64 bits value associated with this WR. If a Work Completion will be generated
    /// when this Work Request ends, it will contain this value.
    ///
    /// See also [RDMAmojo's `ibv_post_send` documentation][1].
    ///
    /// # Safety
    ///
    /// The memory region can only be safely reused or dropped after the request is fully executed
    /// and a work completion has been retrieved from the corresponding completion queue (i.e.,
    /// until `CompletionQueue::poll` returns a completion for this send).
    ///
    /// # Errors
    ///
    ///  - `EINVAL`: Invalid value provided in the Work Request.
    ///  - `ENOMEM`: Send Queue is full or not enough resources to complete this operation.
    ///  - `EFAULT`: Invalid value provided in `QueuePair`.
    ///
    /// [1]: http://www.rdmamojo.com/2013/01/26/ibv_post_send/
```

and refer to the `post` function from this function (same for `write`, `read`)

xmakro (commented):
Thank you! Final set of nits, then we can merge

ibverbs/examples/doorbell_batching.rs
aryeyur: done

ibverbs/src/lib.rs
aryeyur: done

ibverbs/examples/chained_ops.rs
aryeyur: done

ibverbs/examples/chained_ops.rs
aryeyur: done

ibverbs/src/lib.rs
aryeyur: done (in all the methods where it was inaccurate)

aryeyur: done

xmakro:
LGTM, thank you very much for the great contribution!

PR #154, opened 2026-09-17
Rust edition 2024 w/ MSRV 1.85

kykosic:
DWISOTT

Tested compile with Cargo and Bazel, and e2e against a live crate


xmakro:
Thanks!

